PR #33, opened 2019-07-12
Admin section

Joe-LaRue:
Added an admin page that displays users and roles.  currently you can add a user to a role or remove a user from a role.  The admin page and controller are only accessible to those in the administrators role.


src/DevBetterWeb.Web/Views/Shared/_Layout.cshtml
```diff
@@ -27,46 +27,39 @@
...
-                    <li><a asp-area="" asp-page="/Questions">Questions</a></li>
-                </ul>
-            </div>
+            <a class="navbar-brand" href="/">
+                <img src="/images/logo_218x48.png" id="logo" alt="devBetter logo" />
+                <img src="/images/icon_50x50.png" id="icon" alt="devBetter icon" />
+            </a>
+            <a asp-page="/ArchivedVideos/Index">Videos</a>
+            <a asp-page="/Questions">Questions</a>
+            @if (User.IsInRole(Constants.Roles.ADMINISTRATORS))
```
ardalis: These View sections could be replaced with a Tag Helper like this one:
https://www.davepaquette.com/archive/2017/11/05/authorize-tag-helper.aspx

No need to do it for this PR but something to think about for another pass.

PR #40, opened 2019-10-05
100519 admin page cleanup

Joe-LaRue:
When the admin section was originally created, it was done with a controller and views.  I didn't realize the application was using razor pages.  

I recreated the admin section in razor pages, and enhanced a couple of things, including issue #35 

- edit user page
  - you can click on a role to view that role
  - when you add or remove a role, you will remain on that user page, rather than be redirected to admin/index
- edit role page
  - you can click on a user to view that user
  - when you add or remove a user, you will remain on that role page, rather than be redirected to admin/index


ardalis:
Thanks!

PR #64, opened 2020-02-18
Jlarue/expand user data

Joe-LaRue:
added member identity and some pages to support viewing and editing that data.  There is a DB migration that needs to be applied with this code.


ardalis:
I'll need to publish this manually since I don't have the github action pointing at the right resource group yet.

Issue #58, opened 2020-02-07
Show only users in no roles in some view


ardalis:
On the admin users page, be able to filter to see users who aren't in any role.

ardalis:
I need to test and see if this is fixed now by data tables

ardalis:
Sorting the datatable works for this.

PR #68, opened 2020-03-03
Jlarue/add data tables

Joe-LaRue:
adding searchable grids via dataTables to make user management easier!


Joe-LaRue:
related to issue #58 possibly

ardalis:
I’m going to merge a big update tonight that you may need to work into this. Sorry...

ardalis:
Looks good! It won't surprise me if we move to Blazor for most of our rich client needs but this should improve things in the meantime.

Issue #74, opened 2020-03-13
Broken Redirect to User Details


ardalis:
From devbetter.com/User where it lists all Members, if you click View it sends you here:

https://devbetter.com/Home/Index/60b5123e-5392-4b9a-b3bc-fd4a215d4252?page=%2FUser%2FDetails

Which isn't right.

rickhodder:
figured it out - will be checking in a pr in a moment

PR #78, opened 2020-03-14
Fix for Broken Redirect to User Details #74

rickhodder:
Adjusted page directive to point at correct route and name of route parameter to match parameter in page directive route parameter in details.


rickhodder:
turns out I didnt need to change the asp-page directive from "Details" to "/User/Details", only needed to change the  - do you want to reject it and have me resubmit the PR with the fix? 

<td><a class="btn btn-primary btn-sm" asp-page="Details" asp-route-userid="@userKeyValuePair.Key">View</a></td>

ardalis:
Just add a new commit to this branch.

PR #81, opened 2020-03-15
Fix user details redirect

rickhodder:


ardalis:
I think I already fixed this in another PR, didn't I?

rickhodder:
In the pr your merged in, I had added /user/details which worked but Details would be better, and you said check the change in. So i checked it in and submitted it as a pr, maybe i needed to do that under another issue? New to prs

ardalis:
Yeah, it'll be cleaner since I already merged that PR in if you get latest master and branch from there and submit a new PR. Thanks!

Issue #96, opened 2020-04-10
Role DropDown Fails in Production


ardalis:
Gives this error in console:
Uncaught TypeError: Bootstrap's dropdowns require Popper.js (https://popper.js.org/)
    at c.t.toggle (bootstrap.min.js:6)
    at HTMLButtonElement.<anonymous> (bootstrap.min.js:6)
    at Function.each (jquery-3.2.1.min.js:2)
    at r.fn.init.each (jquery-3.2.1.min.js:2)
    at r.fn.init.c._jQueryInterface [as dropdown] (bootstrap.min.js:6)
    at HTMLButtonElement.<anonymous> (bootstrap.min.js:6)
    at HTMLDocument.dispatch (jquery-3.2.1.min.js:3)
    at HTMLDocument.q.handle (jquery-3.2.1.min.js:3)

ardalis:
@Joe-LaRue

ardalis:
Merged your PR but error remains.

ardalis:
I can see that it's pulling in the .js file.
![image](https://user-images.githubusercontent.com/782127/79013716-53594b00-7b37-11ea-9b09-d455f1820113.png)

But still errors:
![image](https://user-images.githubusercontent.com/782127/79013744-5eac7680-7b37-11ea-8d70-39de4c696a01.png)

Joe-LaRue:
popper needs to be included BEFORE bootstrap - my mistake!  will fix and resubmit

ardalis:
That works!

PR #98, opened 2020-04-10
Update _Layout.cshtml with popperJs script reference

Joe-LaRue:
added popperJs script reference to staging and production environments.  fix for #96


ardalis:
Thanks!

Issue #70, opened 2020-03-03
Make header menu responsive


evergreenuni:
The header menu is not very mobile friendly, and the github banner also gets in the way.  Redesign it to make it prettier / more responsive

ardalis:
Thanks!

PR #100, opened 2020-04-10
Nav menu updates

Joe-LaRue:
-removed octo cat from corner and put a link in the nav bar
-created submenu for user account actions
-added responsive bootstrap classes so it will collapse on smaller screens
-reorganized nav items
all for issue  #70 !

### DIFF
--- a/src/DevBetterWeb.Web/Views/Home/Index.cshtml
+++ b/src/DevBetterWeb.Web/Views/Home/Index.cshtml
@@ -1,39 +1,8 @@
 ﻿@{
     ViewData["Title"] = "Developer Career Coaching";
 }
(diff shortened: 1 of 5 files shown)


ardalis:
Wait a minute - I liked the Octocat...

Joe-LaRue:
Do you want to keep the octo cat in the the corner?  Or just somewhere in the page?  

Maybe we can control the visibility of the corner thing with a media query?  Or maybe there is another place we could put it!

ardalis:
I'd like it in the corner but only at PC widths. It can be disappeared via media query on phones, etc. Also the main page already is centered with margins. The cat can disappear any time the margins disappear if that helps your layout.

PR #106, opened 2020-07-27
Initial functionality review

samuelwine:
This is not quite ready yet, as it only has functionality to display a list of subscriptions (and also not in the format of the requirements), but I just thought to ask if you could look over what I've done so far to advise if I'm going in the right direction (or not :-) ) Thank you!


src/DevBetterWeb.Web/Pages/Admin/User.cshtml.cs
```diff
@@ -20,20 +21,23 @@ public class UserModel : PageModel
         private readonly UserManager<ApplicationUser> _userManager;
         private readonly RoleManager<IdentityRole> _roleManager;
         private readonly IUserRoleMembershipService _userRoleMembershipService;
+        private readonly AppDbContext _context;
```
ardalis, line 24: We don't use anything from Infrastructure in Web. Instead we would use an abstraction, like `IRepository`.

If you see `Infrastructure` in the list of using statements for a page, you know there is a problem.

src/DevBetterWeb.Web/Pages/Admin/User.cshtml.cs
```diff
@@ -70,6 +74,21 @@ public async Task<IActionResult> OnGetAsync(string userId)
             RolesNotAssignedToUser = unassignedRoles.Select(x => new SelectListItem(x.Name, x.Id)).ToList();
             Roles = assignedRoles.ToList();
 
+            var subscriptions = await _context.Subscriptions
```
ardalis, line 77: this would be something like:
```
var memberSpecification = new SubscriptionsByMemberSpec(userId);
await _repository.List(memberSpecification);
```
Example spec:
https://github.com/DevBetterCom/DevBetterWeb/blob/master/src/DevBetterWeb.Core/Specs/MemberByUserIdSpec.cs

Example repository usage:
https://github.com/DevBetterCom/DevBetterWeb/blob/master/src/DevBetterWeb.Web/Pages/User/MyProfile.cshtml.cs#L62-L63

ardalis: The new spec would go in the Core/Specs folder and would include the Where and OrderBy bits.

src/DevBetterWeb.Core/Entities/Subscription.cs
```diff
@@ -0,0 +1,14 @@
...
+using System;
+
+namespace DevBetterWeb.Core.Entities
+{
+    public class Subscription : BaseEntity
+    {
+        public DateTime StartDate { get; set; }
+        public DateTime EndDate { get; set; }
+#pragma warning disable CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
+        public Member Member { get; set; }
```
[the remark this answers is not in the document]
samuelwine: thanks for this, will make the changes.
What is the right way to deal with the previous migrations, do i remove them and then create a new one, or just add a new migration on top of the old ones?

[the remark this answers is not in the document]
ardalis: Delete and redo works fine.

ardalis:
@samuelwine do you think you'll have time to look at this soon?

src/DevBetterWeb.Core/Entities/Subscription.cs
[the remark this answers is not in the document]
samuelwine: changes made

src/DevBetterWeb.Web/Pages/Admin/User.cshtml.cs
samuelwine: changes made

src/DevBetterWeb.Web/Pages/Admin/User.cshtml.cs
samuelwine: done

samuelwine:
I've done this tonight - sorry for the e x t e n d e d delay

ardalis:
Great, thanks!

ardalis:
If you think this is ready to merge I'll try and review it closely this week and publish it to the site. Thanks!

samuelwine:
It's not quite ready yet - I'm still in the middle of getting the Add Subscription modal together, and I also still need to implement the checking logic that ensures that the dates are in the correct order (thanks for your response regarding that) and that subscriptions don't overlap.

ardalis:
Ok, thanks. Let me know.

ardalis:
So... ready now? :)

samuelwine:
not quite...
i'm getting there - i've done the edit and delete subscription modals and actions.
tonight i worked on the DateTimeRange value type and how to get that to work with EF - that's done now.
still need to implement:
- exception handling for EndDates prior to StartDate (and display to user); I looked into this tonight and it seems fairly straightforward
- ensuring that subscriptions don't overlap
- display total days from all subscriptions
I probably will manage to complete all this by around Sunday next week

samuelwine:
I got a bit stuck last night trying to add an error to ModelState and then return the page, as the returned page doesn't have all the data it needs (which is set in the OnGet action). 
I thought of two options:
1. return all the necessary properties in the request and use BindProperties attribute
2. add an optional 'ModelErrors' parameter to OnGet and use RedirectToPage passing the error as a parameter
Are either of these a good way, or is there a better way?
Thank you!

ardalis:
For your question I think I would need to see the problem to give a good answer. Usually when you get model validation errors it's in response to a POST, so putting something on OnGet doesn't make sense to me. And the POST won't return its normal result, it will return a BadRequest(modelErrors) so your actual return type shouldn't matter because it's not used.

samuelwine:
I'm getting this error on the build
image
I've updated my nuget packages to the same as this github repo but still getting this error (and now locally as well)
Has there been a change to Ardalis.Specification package that it no longer has BaseSpecification<T>?

samuelwine:
sorted now - i updated the code to fit with the new package versions

samuelwine:
This is ready for review now.
All the functionality detailed in the requirements has been implemented.
I'm unhappy though with how the errors from overlapping and invalid subscriptions are returned to the user (in OnPostAddSubscriptionAsync), if you could advise on the right way to do this I'd be grateful.
Many thanks

ardalis (approved):
I think the roll-up of errors works.

ardalis:
I'll try to get this merged this week, along with the other outstanding PRs. They all require migrations in prod which is what I'm waiting to have some time to do since I expect it will break the site for a bit.

PR #124, opened 2020-08-24
Member update profile discord event

### LINKED ISSUE #122: Add an Event when a Member updates their Profile
Right about here, add an event:
https://github.com/DevBetterCom/DevBetterWeb/blob/master/src/DevBetterWeb.Web/Pages/User/MyProfile.cshtml.cs#L71

And then handle it with a DiscordHandler like this one:
https://github.com/DevBetterCom/DevBetterWeb/blob/master/src/DevBetterWeb.Infrastructure/Handlers/DiscordLogNewUserRegisteredHandler.cs

Message should be something like "User {username} just updated their profile. Check it out here: {memberDetailsUrl}"

MemberDetailsUrl looks something like this:
https://devbetter.com/User/Details/60b5123e-5392-4b9a-b3bc-fd4a215d4252

ilyanaDev:
fixes #122


tests/DevBetterWeb.Tests/Core/Entities/MemberTests/DiscordLogUserUpdateHandlerShould.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+
+        [Fact]
+        public void ReturnsProperMessageString()
+        {
+            Member member = MemberHelpers.CreateWithDefaultConstructor();
+            MemberUpdatedEvent memberEvent = new MemberUpdatedEvent(member, "Links");
+
+            var output = DiscordLogUserUpdateHandler.returnWebhookMessageString(memberEvent);
+            var expected = $"User {memberEvent.Member.FirstName} {memberEvent.Member.LastName} just updated their profile. Check it out here: https://devbetter.com/User/Details/{memberEvent.Member.Id}.";
```
ardalis, line 22: It's best if your expectations

- Don't use the same logic as your system under test
- Are easily readable (as literal values) rather than computed, if possible

Can you change this line to literally just be the expected string?

PR #133, opened 2020-10-05
Books Leader Board PR 1

ilyanaDev:
Members can add and remove books from their Books Read list from dropdown menus in My Profile
Admins have CRUD capability on Books

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,123 @@
+# To learn more about .editorconfig see https://aka.ms/editorconfigdocs
+###############################
+# Core EditorConfig Options   #
+###############################
+# All files
+[*]
+indent_style = space
+# Code files
+[*.{cs,csx,vb,vbx}]
+indent_size = 2
+insert_final_newline = true
+charset = utf-8-bom
+###############################
+# .NET Coding Conventions     #
+###############################
+[*.{cs,vb}]
+# Organize usings
+dotnet_sort_system_directives_first = true
+# this. preferences
+dotnet_style_qualification_for_field = false:silent
+dotnet_style_qualification_for_property = false:silent
+dotnet_style_qualification_for_method = false:silent
+dotnet_style_qualification_for_event = false:silent
+# Language keywords vs BCL types preferences
+dotnet_style_predefined_type_for_locals_parameters_members = true:silent
+dotnet_style_predefined_type_for_member_access = true:silent
+# Parentheses preferences
+dotnet_style_parentheses_in_arithmetic_binary_operators = always_for_clarity:silent
+dotnet_style_parentheses_in_relational_binary_operators = always_for_clarity:silent
+dotnet_style_parentheses_in_other_binary_operators = always_for_clarity:silent
+dotnet_style_parentheses_in_other_operators = never_if_unnecessary:silent
+# Modifier preferences
+dotnet_style_require_accessibility_modifiers = for_non_interface_members:silent
+dotnet_style_readonly_field = true:suggestion
+# Expression-level preferences
+dotnet_style_object_initializer = true:suggestion
+dotnet_style_collection_initializer = true:suggestion
+dotnet_style_explicit_tuple_names = true:suggestion
+dotnet_style_null_propagation = true:suggestion
+dotnet_style_coalesce_expression = true:suggestion
+dotnet_style_prefer_is_null_check_over_reference_equality_method = true:silent
+dotnet_style_prefer_inferred_tuple_names = true:suggestion
+dotnet_style_prefer_inferred_anonymous_type_member_names = true:suggestion
+dotnet_style_prefer_auto_properties = true:silent
+dotnet_style_prefer_conditional_expression_over_assignment = true:silent
+dotnet_style_prefer_conditional_expression_over_return = true:silent
+###############################
+# Naming Conventions          #
+###############################
+# Style Definitions
+dotnet_naming_style.pascal_case_style.capitalization             = pascal_case
+# Use PascalCase for constant fields  
+dotnet_naming_rule.constant_fields_should_be_pascal_case.severity = suggestion
+dotnet_naming_rule.constant_fields_should_be_pascal_case.symbols  = constant_fields
+dotnet_naming_rule.constant_fields_should_be_pascal_case.style    = pascal_case_style
+dotnet_naming_symbols.constant_fields.applicable_kinds            = field
+dotnet_naming_symbols.constant_fields.applicable_accessibilities  = *
+dotnet_naming_symbols.constant_fields.required_modifiers          = const
+###############################
+# C# Coding Conventions       #
+###############################
+[*.cs]
+# var preferences
+csharp_style_var_for_built_in_types = true:silent
+csharp_style_var_when_type_is_apparent = true:silent
+csharp_style_var_elsewhere = true:silent
+# Expression-bodied members
+csharp_style_expression_bodied_methods = false:silent
+csharp_style_expression_bodied_constructors = false:silent
+csharp_style_expression_bodied_operators = false:silent
+csharp_style_expression_bodied_properties = true:silent
+csharp_style_expression_bodied_indexers = true:silent
+csharp_style_expression_bodied_accessors = true:silent
+# Pattern matching preferences
+csharp_style_pattern_matching_over_is_with_cast_check = true:suggestion
+csharp_style_pattern_matching_over_as_with_null_check = true:suggestion
+# Null-checking preferences
+csharp_style_throw_expression = true:suggestion
+csharp_style_conditional_delegate_call = true:suggestion
+# Modifier preferences
+csharp_preferred_modifier_order = public,private,protected,internal,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,volatile,async:suggestion
+# Expression-level preferences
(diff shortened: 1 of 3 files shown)


src/DevBetterWeb.Core/Entities/BookMember.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace DevBetterWeb.Core.Entities
```
ardalis, line 1: Remove this file.

src/DevBetterWeb.Core/JoinCollectionFacade.cs
```diff
@@ -0,0 +1,67 @@
+﻿using DevBetterWeb.Core.SharedKernel;
```
ardalis, line 1: Remove this file.

src/DevBetterWeb.Core/SharedKernel/ValueObject.cs
```diff
@@ -6,108 +6,108 @@
 namespace DevBetterWeb.Core.SharedKernel
```
ardalis, line 6: Remove file.

src/DevBetterWeb.Core/Specs/MemberByUserIdWithBooksReadSpec.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace DevBetterWeb.Core.Specs
+{
+    public class MemberByUserIdWithBooksReadSpec : Specification<Member>
+    {
+        public string UserId { get; }
+
+        public MemberByUserIdWithBooksReadSpec(string userId)
+        {
+            UserId = userId;
```
ardalis, line 12: What is this property used for?

src/DevBetterWeb.Infrastructure/Data/AppDbContext.cs
```diff
@@ -22,12 +22,16 @@ public class AppDbContext : DbContext
         protected override void OnModelCreating(ModelBuilder modelBuilder)
         {
             base.OnModelCreating(modelBuilder);
+
+            modelBuilder.Entity<Book>().Property(b => b.Title).HasMaxLength(100);
```
ardalis, line 26: This belongs in a config class in the config folder.

ardalis: Actually you already have `/Config/BookConfig.cs` so this should be removed.

src/DevBetterWeb.Infrastructure/Data/Config/BookMemberConfig.cs
```diff
@@ -0,0 +1,23 @@
+﻿using DevBetterWeb.Core.Entities;
```
ardalis, line 1: Remove file.

src/DevBetterWeb.Infrastructure/Data/Migrations/20200929181542_FixBookMember.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
ardalis, line 1: This one can probably be deleted since it does nothing.

src/DevBetterWeb.Web/DevBetterWeb.Web.csproj
```diff
@@ -1,68 +1,71 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
-  <PropertyGroup>
```
ardalis, line 2: See if you can reformat this to match what was there before so every line isn't different due to whitespace. There should be a .editorconfig file which sets spaces per tab to 2. You seem to have use 4 (or more).

src/DevBetterWeb.Web/Pages/User/UserProfileUpdateModel.cs
```diff
@@ -10,28 +14,32 @@ namespace DevBetterWeb.Web.Pages.User
...
     public class UserProfileUpdateModel
     {
+
         [Required]
         public string? FirstName { get; set; }
         [Required]
         public string? LastName { get; set; }
         public string? Address { get; set; }
-        [ValidUrl]
+        //[ValidUrl]
```
ardalis, line 23: Why are we changing ValidUrl? Did it break somehow?

src/DevBetterWeb.Web/Program.cs
```diff
@@ -20,7 +20,7 @@ public static async Task Main(string[] args)
         {
             var builder = CreateHostBuilder(args);
 
-            //string env = "Development";
+            string env = "Development";
```
ardalis, line 23: This probably needs to be the way it was. We don't want to force the environment to always be Development when we deploy to production.

tests/DevBetterWeb.Tests/Core/Entities/MemberTests/BookHelpers.cs
```diff
@@ -0,0 +1,20 @@
+﻿using DevBetterWeb.Core.Entities;
+
+namespace DevBetterWeb.Tests.Core.Entities.MemberTests
+{
+    public static class BookHelpers
+    {
+        public static Book GetDefaultBook()
```
ardalis, line 7: This is good to use for tests instead of newing up a book. I'd make it more clear this is for tests somehow in the name, ideally, like GetDefaultTestBook or GetTestBook etc. Could also be an opportunity to use AutoFixture.
https://github.com/AutoFixture/AutoFixture/wiki/Cheat-Sheet#complex-type

tests/DevBetterWeb.Tests/Core/Entities/MemberTests/MemberAddBook.cs
```diff
@@ -0,0 +1,36 @@
...
+            Member member = MemberHelpers.CreateWithDefaultConstructor();
+            Book book = BookHelpers.GetDefaultBook();
+
+            member.AddBookRead(book);
+            member.AddBookRead(book);
+
+            // if we get to this point no error was thrown.
+            Assert.Contains(book, member.BooksRead);
+            Assert.Single(member.BooksRead!);
+        }
```
ardalis, line 33: Good tests!

tests/DevBetterWeb.Tests/Core/Entities/MemberTests/UserProfileUpdateModelHasReadBook.cs
```diff
@@ -0,0 +1,70 @@
...
+            bool hasReadBook = updateModel.HasReadBook(Book);
+
+            Assert.True(hasReadBook);
+        }
+
+        [Fact]
+        public void ShouldReturnTrueIfBooksReadContainsBookAndAnotherBook()
+        {
+            UserProfileUpdateModel updateModel = GetUserProfileUpdateModelWithBookInBooksRead();
+            updateModel.BooksRead.Add(new DevBetterWeb.Core.Entities.Book { Author = "Steve Smith", Title = "C# stuff", Details = "A book about c#", PurchaseUrl = "https://buyabook.com" });
```
ardalis, line 50: Would be good to stick with your BookHelpers to do this. Perhaps have an optional parameter to specify a different title, which would be sufficient to give you a different book, while still keeping `new` out of your tests.

ardalis (changes requested):

src/DevBetterWeb.Core/Specs/MemberByUserIdWithBooksReadSpec.cs
ilyanaDev: it's used to make sure you get the member with the userId you pass in

src/DevBetterWeb.Web/Pages/User/UserProfileUpdateModel.cs
ilyanaDev: it was broken at one point, while I had a work-in-progress. Should be ok to change it back now. I will try

ilyanaDev: it was broken because of some of the changes we made with nullability

src/DevBetterWeb.Core/Specs/MemberByUserIdWithBooksReadSpec.cs
ilyanaDev: removed

src/DevBetterWeb.Infrastructure/Data/AppDbContext.cs
ilyanaDev: removed

src/DevBetterWeb.Web/DevBetterWeb.Web.csproj
ilyanaDev: this is another thing I didn't change. Seems to me it is properly nested. (PropertyGroup is inside Project and thus _should_ be indented farther, and that holds true for the whole file)

src/DevBetterWeb.Web/Pages/User/UserProfileUpdateModel.cs
ilyanaDev: so essentially, in ValidUrlAttribute.cs, the overriding IsValid method is mad about the nullability of its parameter value, because we enabled nullability in the project. Can I suppress that error for that line of code?

PR #141, opened 2020-10-16
Codin game and url verification

ilyanaDev:


ardalis:
ignoring build failure due to .net 5 not 3.1

PR #170, opened 2020-12-25
Issue 107 - Map with Pins

samuelwine:
This PR covers the basic map functionality with markers for each member. 
It does not include code for the map marker icons on the members page.


samuelwine:
I've been thinking about this a bit - it really still exposes member addresses as the pins are pinpointed to the actual address if you zoom in. 
Would a sensible approach be to hit the GoogleMaps API twice, the first time with the address from the DB, then extract the city co-ordinates from the response, then hit the API again with just the city co-ordinates - so the pin will be generally centered over the city?

ardalis:
Can we hit the API in order to get the location of just the city? We don't need street level detail anywhere in this operation, I think. If we are able to fetch the Lat/Long of the city, we really only need that once. We could store it in the database with the user's details and only update it if they change their address. Thoughts?

samuelwine:
If we could extract the city from the address string, that would be no problem to do. Since we are storing the address as a comma seperated string, could we be sure that the city (or general area) would be between the second and third comma?

ardalis:
Let's try that and if it doesn't work I'll just ask for Address differently with proper breakdown of its elements. When all I needed it for was shipping, it was fine to just have a string. But if we want it for this purpose as well, then exposing the City/State/Country for map pins probably makes sense.

samuelwine:
I've just updated the PR to use just the city and not the full address.
Next step is to save the API response with the user details as you mentioned above.
How do I know if the address has changed to know to fetch the new map details - would that be with a domain event raised on address change?

ardalis:
Yes, I think that makes the most sense.

samuelwine:
This is completed and ready for review - thanks!

ardalis:
Nice. I'll try and review and merge it today!

samuelwine:
Thanks!

Any feedback would be really appreciated - even if something is "ok" but
you would do it in a better way, it would be really useful for me to know!

src/DevBetterWeb.Infrastructure/Data/AppDbContext.cs
```diff
@@ -34,6 +36,31 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
         public override async Task<int> SaveChangesAsync(CancellationToken cancellationToken = new CancellationToken())
         {
+            if (_dispatcher is not null)
```
ardalis, line 39: Why are we adding this code here when there is already code to handle every kind of domain event after the call to `base.SaveChangesAsync()` ?

src/DevBetterWeb.Infrastructure/Handlers/MemberAddressUpdatedHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.Extensions.Configuration;
+using Newtonsoft.Json.Linq;
+
+namespace DevBetterWeb.Infrastructure.Handlers
+{
+  class MemberAddressUpdatedHandler : IHandle<MemberAddressUpdatedEvent>
+  {
+    public MemberAddressUpdatedHandler(IConfiguration configuration)
+    {
+      _configuration = configuration;
```
ardalis, line 18: This handler should depend on a service for getting coordinates, not the config for the google key. Also the _configuration property should probably be a private field (if we were keeping it here).

What I would recommend instead is:
- Create an interface in Core/Interfaces for something like IMapCoordinateService with method `Task<string> GetMapCoordinates(string address)`
- Create a service in Infrastructure/Services called GoogleMapCoordinateService : IMapCoordinateService and put the private method you have in this class into the service. Inject IConfiguration into the service.
- Inject the `IMapCoordinateService` into this handler and just call it.
- Consider injecting an `HttpClient` into the MapCoordinateService rather than newing it up in a using block. I think this will be better but I'm not 100% sure since the guidance on this has changed back and forth a bit.

src/DevBetterWeb.Infrastructure/Handlers/MemberAddressUpdatedHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+    public IConfiguration _configuration { get; }
+
+    public async Task Handle(MemberAddressUpdatedEvent addressUpdatedEvent)
+    {
+      var member = addressUpdatedEvent.Member;
+      var addressParts = member.Address?.Split(',');
+
+      if (addressParts is not null)
+      {
+        var cityDetailsAPIResponse = JObject.Parse(await GetMapCoordinates(addressParts[1]));
```
ardalis, line 30: Parsing the address should be its own service. If `GetMapCoordinates` needs a particular part of an address, not a full address string, then rename its `address` parameter to something like `cityCountryString` or similar.

src/DevBetterWeb.Infrastructure/Handlers/MemberAddressUpdatedHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+        var latResponse = cityDetailsAPIResponse.SelectToken("results[0].geometry.location.lat");
+        var lngResponse = cityDetailsAPIResponse.SelectToken("results[0].geometry.location.lng");
+
+        member.CityLatitude = latResponse?.ToObject<decimal>();
+        member.CityLongitude = lngResponse?.ToObject<decimal>();
+      }
```
ardalis, line 36: I'm guessing the reason you moved the code in the `SaveChanges` method is to make sure this handler gets called before changes are saved, so that the CityLatitude and CityLongitude are also saved. Your approach works, but it's not terribly clean. We should think about a better approach, such as implementing an immediate domain event pattern (which has its own complexities, to be sure). Another option would be to have some totally out-of-proc service be responsible for the lat/long updates, and this handler would just put a command to do the check on a queue which the other process would handle, but that's obviously more complex, too. Let's discuss - I'm leaning toward adding immediate domain events to the solution and using them for this.

src/DevBetterWeb.Infrastructure/Handlers/MemberAddressUpdatedHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    private async Task<string> GetMapCoordinates(string address)
```
ardalis, line 39: Generally handlers shouldn't have private methods, but instead should have services injected into them to do whatever work they require.

ardalis (changes requested):
Please refactor the handler to use an injected service instead of a private method.
Let's discuss how to implement immediate domain events in the solution. We might merge your existing changes (since I think they work) and then refactor after.

src/DevBetterWeb.Infrastructure/Handlers/MemberAddressUpdatedHandler.cs
samuelwine: This is done. 
Thank you for giving me the steps.

src/DevBetterWeb.Infrastructure/Handlers/MemberAddressUpdatedHandler.cs
samuelwine: I've renamed the  parameter, but I wasn't clear what you meant with "Parsing the address should be its own service"?
Thanks

src/DevBetterWeb.Infrastructure/Handlers/MemberAddressUpdatedHandler.cs
samuelwine: This has been changed as per the steps you laid out

PR #206, opened 2021-02-17
Stripe signup flow in progress

ilyanaDev:
Most of issue 204 is fulfilled here. NewMemberService and a StripeWebhookHandler are created to handle the signup flow as new members sign up.


efscripts.txt
```diff
@@ -1,6 +1,6 @@
 
 -- RUN THIS FROM THE WEB PROJECT FOLDER
-dotnet ef migrations add DiscordUsername --context appdbcontext -p ../DevBetterWeb.Infrastructure/DevBetterWeb.Infrastructure.csproj -s DevBetterWeb.Web.csproj -o Data/Migrations
+dotnet ef migrations add InviteEntityActiveBool --context appdbcontext -p ../DevBetterWeb.Infrastructure/DevBetterWeb.Infrastructure.csproj -s DevBetterWeb.Web.csproj -o Data/Migrations
```
ardalis, line 3: We don't need to update this every time we create a new migration - it's meant to be a template with the name as a placeholder to be replaced (in the console, not here).

src/DevBetterWeb.Core/Interfaces/INewMemberService.cs
```diff
@@ -0,0 +1,14 @@
+﻿using DevBetterWeb.Core.Entities;
+using System.Threading.Tasks;
+
+namespace DevBetterWeb.Core.Interfaces
+{
+  public interface INewMemberService
+  {
+    Task<Invitation> CreateInvitation(string email, string stripeEventId);
```
ardalis, line 8: All async methods should end in `Async` in their names.

src/DevBetterWeb.Core/Interfaces/IPaymentHandlerSubscription.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace DevBetterWeb.Core.Interfaces
+{
+  public interface IPaymentHandlerSubscription
+  {
+    DateTime GetStartDate(string subscriptionId);
+    DateTime GetEndDate(string subscriptionId);
```
ardalis, line 8: Will these frequently be called one after the other by whatever client code uses this interface? Is there a way to combine them into one method that returns a date range type or Tuple?

src/DevBetterWeb.Core/Interfaces/IUserRoleMembershipService.cs
```diff
@@ -2,9 +2,10 @@
...
 {
-    public interface IUserRoleMembershipService
-    {
-        Task AddUserToRoleAsync(string userId, string roleId);
-        Task RemoveUserFromRoleAsync(string userId, string roleId);
+  public interface IUserRoleMembershipService
+  {
+    Task AddUserToRoleAsync(string userId, string roleId);
+    Task RemoveUserFromRoleAsync(string userId, string roleId);
+    Task AddUserToRoleAsyncByRoleName(string userId, string roleName);
```
ardalis, line 9: Move `Async` to end of name, not in the middle.

src/DevBetterWeb.Core/Services/NewMemberService.cs
```diff
@@ -0,0 +1,140 @@
...
+      string code = invitation.InviteCode;
+      string inviteEmail = invitation.Email;
+
+      string completeRegistrationUrl = GetRegistrationUrl(code, inviteEmail);
+
+      // TODO: send Discord invite here as well
+
+      var message = "Thank you for joining DevBetter! Please click to complete your registration: " + completeRegistrationUrl + "\n";
+
+      await _emailService.SendEmailAsync(inviteEmail, "Welcome to DevBetter!", message);
```
ardalis, line 51: I suspect this method doesn't actually need to be async. Visual Studio will usually tell you if it's not required. Since there is only one await and nothing depends on the result I think you probably can return Task but not have the `async` or `await` keywords. See if Visual Studio underlines the method as it is now.

src/DevBetterWeb.Core/Services/NewMemberService.cs
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    public async Task<string> VerifyValidEmailAndInviteCode(string email, string inviteCode)
+    {
+      var spec = new InvitationByInviteCodeWithEmailSpec(inviteCode);
+
+      string ValidEmailAndInviteCode = "success";
```
ardalis, line 58: Think about replacing this string-based result with Ardalis.Result nuget package.

src/DevBetterWeb.Core/Services/NewMemberService.cs
```diff
@@ -0,0 +1,140 @@
...
+        ValidEmailAndInviteCode = "Invalid email or invite code: " + e.GetType().ToString();
+      }
+
+      return ValidEmailAndInviteCode;
+    }
+
+    public async Task<Member> MemberSetup(string userId, string firstName, string lastName, string inviteCode)
+    {
+      Member member = CreateNewMember(userId, firstName, lastName);
+      int memberId = member.Id;
```
ardalis, line 87: Just use member.Id below in the one place it's needed.

src/DevBetterWeb.Core/Services/NewMemberService.cs
```diff
@@ -0,0 +1,140 @@
...
+    {
+      Member member = new Member(userId);
+      member.UpdateName(firstName, lastName);
+
+      return member;
+    }
+
+    private async Task AddUserToMemberRole(string userId)
+    {
+      var roleName = "Members";
```
ardalis, line 117: This should be a constant, which probably already exists somewhere but if not, add it.

src/DevBetterWeb.Core/Services/NewMemberService.cs
```diff
@@ -0,0 +1,140 @@
...
+      var subscription = new Subscription();
+      subscription.MemberId = memberId;
+      DateTimeRange dates = new DateTimeRange(subscriptionStart, subscriptionEnd);
+      subscription.Dates = dates;
+      return subscription;
+    }
+
+    private string GetRegistrationUrl(string inviteCode, string inviteEmail)
+    {
+      var url = $"https://devbetter.com/Identity/Account/NewMemberRegister/{inviteCode}/{inviteEmail}";
```
ardalis, line 134: Careful using the full domain here since it will make it harder to test or use in different environments or use in a container, etc. Usually you can have the domain passed in or read from configuration. This is fine for now just FYI.

src/DevBetterWeb.Core/Specs/InvitationByInviteCodeWithEmailSpec.cs
```diff
@@ -0,0 +1,14 @@
...
+using DevBetterWeb.Core.Entities;
+
+namespace DevBetterWeb.Core.Specs
+{
+  public class InvitationByInviteCodeWithEmailSpec : Specification<Invitation>
+  {
+    public InvitationByInviteCodeWithEmailSpec(string inviteCode)
+    {
+      Query.Where(invite => invite.InviteCode == inviteCode);
+      Query.Include(invite => invite.Email);
```
ardalis, line 11: If Email is a string property, no need to .Include it.

src/DevBetterWeb.Core/Specs/InvitationByInviteCodeWithSubscriptionIdSpec.cs
```diff
@@ -0,0 +1,14 @@
...
+using DevBetterWeb.Core.Entities;
+
+namespace DevBetterWeb.Core.Specs
+{
+  public class InvitationByInviteCodeWithSubscriptionIdSpec : Specification<Invitation>
+  {
+    public InvitationByInviteCodeWithSubscriptionIdSpec(string inviteCode)
+    {
+      Query.Where(invite => invite.InviteCode == inviteCode);
+      Query.Include(invite => invite.PaymentHandlerSubscriptionId);
```
ardalis, line 11: No need to .Include primitive types on the entity. Just navigation properties to other entities.

src/DevBetterWeb.Infrastructure/Data/Migrations/20210209190217_InviteEntity.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            migrationBuilder.CreateTable(
+                name: "Invitations",
+                columns: table => new
+                {
+                    Id = table.Column<int>(type: "int", nullable: false)
+                        .Annotation("SqlServer:Identity", "1, 1"),
+                    Email = table.Column<string>(type: "nvarchar(max)", nullable: true),
+                    InviteCode = table.Column<string>(type: "nvarchar(max)", nullable: true),
+                    StripeSubscriptionId = table.Column<string>(type: "nvarchar(max)", nullable: true)
```
ardalis, line 17: Add an InviteConfig file in /Data/Config that sets the MaxLength on the string properties to something reasonable. There may be constants available for common things like Email.

src/DevBetterWeb.Infrastructure/PaymentHandler/StripePaymentHandler/StripeCustomer.cs
```diff
@@ -0,0 +1,28 @@
...
+
+      var email = customer.Email;
+
+      return email;
+
+    }
+
+    private Stripe.Customer GetCustomer(string customerId)
+    {
+      var customerService = new CustomerService();
```
ardalis, line 20: This should be injected in the constructor.

src/DevBetterWeb.Infrastructure/PaymentHandler/StripePaymentHandler/StripeCustomer.cs
```diff
@@ -0,0 +1,28 @@
+﻿using DevBetterWeb.Core.Interfaces;
+using Stripe;
+
+namespace DevBetterWeb.Infrastructure.PaymentHandler.StripePaymentHandler
+{
+  public class StripeCustomer : IPaymentHandlerCustomer
```
ardalis, line 6: This is a service or a handler (which is a special kind of service), so its name should have a suffix reflecting this.

src/DevBetterWeb.Infrastructure/PaymentHandler/StripePaymentHandler/StripeSubscription.cs
```diff
@@ -0,0 +1,39 @@
...
+      var subscription = GetSubscription(subscriptionId);
+
+      DateTime startDate = subscription.CurrentPeriodStart;
+
+      return startDate;
+    }
+
+    private Stripe.Subscription GetSubscription(string subscriptionId)
+    {
+      var subscriptionService = new SubscriptionService();
```
ardalis, line 30: Inject

src/DevBetterWeb.Core/Services/NewMemberService.cs
ilyanaDev: it doesn't really need to be it seems like. But given that it does call an async method, would it be better to make it clear that the email _is_ sent async by keeping this method async?

PR #207, opened 2021-03-08
Stripe signup flow

### LINKED ISSUE #173: Add sign up functionality
This should handle both monthly and yearly subscriptions, as well as free sign ups. It will need to be able to integrate with Stripe. It should also have the functionality to cancel and refund signups.

ilyanaDev:
Stripe signup flow should now work.

Fixes #173


ardalis:
Does this make the feature public in any way, or is it still hidden but now theoretically works?

ilyanaDev:
all the pages should be admin only. But I will double check that.

src/DevBetterWeb.Core/Services/NewMemberService.cs
```diff
@@ -47,7 +47,7 @@ public async Task SendRegistrationEmailAsync(Invitation invitation)
 
       // TODO: send Discord invite here as well
 
-      var message = "Thank you for joining DevBetter! Please click to complete your registration: " + completeRegistrationUrl + "\n";
+      var message = "Thank you for joining DevBetter! Please click to complete your registration: " + completeRegistrationUrl + "\n\nWe're so glad to have you here!";
```
ardalis, line 50:
```suggestion
      var message = $"Thank you for joining DevBetter! Please click to complete your registration:\n\n {completeRegistrationUrl}\n\nWe're so glad to have you here!";
```

src/DevBetterWeb.Web/Controllers/BillingController.cs
```diff
@@ -65,7 +75,22 @@ public ActionResult<Subscription> CreateSubscription(SubscriptionCreateRequest r
       }
       catch (StripeException e)
       {
-        return BadRequest(e);
+        var error = new SubscriptionError(e.Message);
+
+        return error;
+      }
```
ardalis, line 81: Since both of these catch blocks do the same thing, is there any reason the try block can't be made bigger, so there aren't duplicate catch blocks?

ilyanaDev:
all the new pages are still restricted to admins

src/DevBetterWeb.Web/Controllers/StripeWebhookHandler.cs
```diff
@@ -39,27 +40,37 @@ public class StripeWebhookHandler : Controller
...
-
-      return Ok();
 
       try
       {
         var stripeEvent = EventUtility.ParseEvent(json);
         var stripeEventType = stripeEvent.Type;
 
-        if (stripeEventType.Equals("customer.subscription.created"))
+        if (stripeEventType.Equals("customer.subscription.created") || stripeEventType.Equals("customer.subscription.updated"))
```
ardalis, line 49: Use constants for these strings so they are only defined in one place in the entire source code.

ardalis (changes requested):
Looks good - a few minor suggestions. Also try to deal with the compiler warnings.

PR #211, opened 2021-03-22
Stripe adapter

ilyanaDev:
all the stripe payment handler code (aside from the code in the js) is now pulled behind a set of interfaces


src/DevBetterWeb.Core/Interfaces/IPaymentHandlerSubscriptionDTO.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace DevBetterWeb.Core.Interfaces
+{
+  public interface IPaymentHandlerSubscriptionDTO
```
ardalis, line 3: DTOs don't typically need interfaces. Is there a reason for this one vs just a class?

src/DevBetterWeb.Infrastructure/DefaultInfrastructureModule.cs
```diff
@@ -46,6 +48,16 @@ private void RegisterCommonDependencies(ContainerBuilder builder)
...
 
+    private void RegisterPaymentHandlerDependencies(ContainerBuilder builder)
+    {
+      builder.RegisterType<CustomerService>();
+      builder.RegisterType<PaymentMethodService>();
+      builder.RegisterType<SubscriptionService>();
+      builder.RegisterType<PriceService>();
+      builder.RegisterType<PaymentIntentService>();
+      builder.RegisterType<PaymentMethodService>();
+    }
```
ardalis, line 59: I like that these are all grouped in one helper method.

ardalis (approved):
Looks good. My only question is why the interface for the DTO. It's a lot of services and maybe we can reorganize to be not quite so many but otherwise it looks good to me.

ilyanaDev:
fixes #211

PR #216, opened 2021-04-12
Subscription and billing details refactoring

ilyanaDev:


src/DevBetterWeb.Core/ValueObjects/BillingDetails.cs
```diff
@@ -1,24 +1,115 @@
...
+          message += $"Member ID: {memberId}. {MemberName}";
+          break;
+        case Viewer.Member:
+          message += $"You";
+          break;
+        default:
+          break;
+      }
+
+      switch (ActionVerbPastTense)
```
ardalis, line 56: Ugh I hate this. Last thing I wanted when introducing the verb variable was to need a switch. I can see why you went this way, though.

I don't have a better solution at the moment. And at least you should be able to write unit tests for this one easily enough.

src/DevBetterWeb.Core/ValueObjects/BillingDetails.cs
```diff
@@ -1,24 +1,115 @@
...
     {
       yield return Amount;
-      yield return Message;
+      yield return SubscriptionPlanName;
+      yield return MemberName;
+      yield return ActionVerbPastTense;
+      yield return BillingPeriod;
+    }
+
+    public BillingDetails()
```
ardalis, line 100: No need to add constructor initialization for default values, and whenever possible prefer setting fields/properties inline (where they are declared) rather than in the constructor.

Save the constructor for:
1) assigning arguments passed into it
2) validation/guard class on such arguments
3) initialization that can't be done inline or as the default for the type

src/DevBetterWeb.Infrastructure/PaymentHandler/StripePaymentHandler/StripePaymentHandlerSubscriptionService.cs
```diff
@@ -97,6 +99,31 @@ public string GetStatus(string subscriptionId)
...
+
+      return billingPeriod;
+    }
+
+    private BillingPeriod GetSubscriptionBillingInterval(Subscription subscription)
+    {
+      var item = subscription.Items.Data[0];
+      var period = item.Price.Recurring.Interval;
+
+      if (period == "month")
```
ardalis, line 116: I think BillingPeriod is a SmartEnum, right?

If so, you can overload its constructor so you don't need this if logic:

```
public BillingPeriod(string stripePeriod)
{
  // put if logic here
 // put new InvalidBillingPeriodException here, too
}
```

src/DevBetterWeb.Infrastructure/PaymentHandler/StripePaymentHandler/StripePaymentHandlerSubscriptionService.cs
```diff
@@ -97,6 +99,31 @@ public string GetStatus(string subscriptionId)
...
+      {
+        return BillingPeriod.Month;
+      }
+      if (period == "year")
+      {
+        return BillingPeriod.Year;
+      }
+      throw new InvalidBillingPeriodException();
```
ardalis, line 124: Then in here you can just `return new BillingPeriod(period);`

src/DevBetterWeb.Infrastructure/PaymentHandler/StripePaymentHandler/StripePaymentHandlerSubscriptionService.cs
```diff
@@ -113,12 +140,20 @@ private DateTime GetStartDate(Subscription subscription)
 
     private Stripe.Subscription GetSubscription(string subscriptionId)
     {
-
       var subscription = _subscriptionService.Get(subscriptionId);
 
       return subscription;
     }
 
+    public string GetAssociatedProductName(string subscriptionId)
```
ardalis, line 148: Why not get the subscription, and then you have its product name with it?

tests/DevBetterWeb.Tests/Core/ValueObjects/BillingDetailsGetMessageForAdminView.cs
```diff
@@ -0,0 +1,56 @@
+﻿using DevBetterWeb.Core.ValueObjects;
+using Xunit;
+using DevBetterWeb.Core.Enums;
+using System;
+
+namespace DevBetterWeb.Tests.Core.ValueObjects
+{
+  public class BillingDetailsGetMessageForAdminView : BillingDetailsGetMessageTest
```
ardalis, line 8: 👍

ardalis (changes requested):

src/DevBetterWeb.Core/Enums/BillingPeriod.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace DevBetterWeb.Core.Enums
+{
+  public enum BillingPeriod 
+  {
+    None,
```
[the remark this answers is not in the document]
ilyanaDev: I added it because I needed a default constructor for billingdetails for ef core

src/DevBetterWeb.Core/Interfaces/IMemberAddBillingActivityService.cs
```diff
@@ -1,12 +1,13 @@
 ﻿using System.Threading.Tasks;
+using DevBetterWeb.Core.Enums;
 
 namespace DevBetterWeb.Core.Interfaces
 {
   public interface IMemberAddBillingActivityService
```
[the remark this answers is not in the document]
ilyanaDev: they're all called by the same class, (WebhookHandlerService), just by different methods in it

src/DevBetterWeb.Core/Interfaces/IPaymentHandlerSubscription.cs
```diff
@@ -14,5 +15,7 @@ public interface IPaymentHandlerSubscription
     IPaymentHandlerSubscriptionDTO CreateSubscriptionError(string errorMessage);
     bool GetCancelAtPeriodEnd(string subscriptionId);
     Task CancelSubscriptionAtPeriodEnd(string customerEmail);
+    BillingPeriod GetBillingPeriod(string subscriptionId);
+    string GetAssociatedProductName(string subscriptionId);
```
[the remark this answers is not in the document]
ilyanaDev: this is the stripe adapter; it needs to get that information from Stripe itself and pass it to places that need it.

src/DevBetterWeb.Infrastructure/PaymentHandler/StripePaymentHandler/StripePaymentHandlerSubscriptionService.cs
ilyanaDev: I had it as a SmartEnum but then I was having trouble putting it in the database and didn't see any real benefit to not just using a regular enum so I switched it to that. I could switch it back to SmartEnum if you think it's worth it for this

src/DevBetterWeb.Infrastructure/PaymentHandler/StripePaymentHandler/StripePaymentHandlerSubscriptionService.cs
ilyanaDev: this method is to get from Stripe the product name to be assigned to the subscriptionPlanName in SubscriptionPlanDetails in SubscriptionPlan, which a DevBetter (Member)Subscription has. I'm getting a stripe subscription, but I only want to return the product name because I don't want to pass a stripe subscription all over the place

src/DevBetterWeb.Core/Interfaces/IMemberSubscriptionPeriodCalculationsService.cs
```diff
@@ -5,6 +5,7 @@ namespace DevBetterWeb.Core.Interfaces
 {
   public interface IMemberSubscriptionPeriodCalculationsService
```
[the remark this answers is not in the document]
ilyanaDev: It's not returning properties, for the most part. It really is doing calculations. A part of it is that it's possible for members to have multiple subscriptions, so I have to determine which is the current subscription if there are multiple. This class lets me do that without cluttering code elsewhere with that logic. There's an argument to be made for these methods belonging in Member, but personally I like it in its own class where it's conceptually consistent and isn't lost in Member's already large interface.

src/DevBetterWeb.Core/Services/MemberSubscriptionPeriodCalculationsService.cs
```diff
@@ -8,6 +8,20 @@ public class MemberSubscriptionPeriodCalculationsService : IMemberSubscriptionPe
   {
     private const int DAYS_SUBSCRIBED_TO_BECOME_ALUMNUS = 730;
 
+    public bool GetHasCurrentSubscription(Member member)
```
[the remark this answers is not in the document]
ilyanaDev: see above. It certainly could be

src/DevBetterWeb.Core/ValueObjects/BillingDetails.cs
ilyanaDev: I need this for EF Core, don't I? Since the properties aren't nullable some default needs to be given

ilyanaDev:
Ah I know why I made Details nullable in SubscriptionPlan. It yells at me about no suitable constructor and blows up if Details isn't nullable

ardalis (approved):

ardalis:
What yells at you if it's not nullable?

When I merge this, will it require db schema updates in prod?

ilyanaDev:
> What yells at you if it's not nullable?
> 
> When I merge this, will it require db schema updates in prod?

it was doing this on the MyProfile page: 
image

I fixed that by adding parameterless constructors for SubscriptionPlan and SubscriptionPlanDetails and now it works without details having to be nullable

Also, yes, it will require database updates because I've added columns to tables already visible in prod, and when I roll the database back to the migration it's at in prod, I get errors.

ardalis:
Merged - let's see if it all works...

PR #265, opened 2021-04-21
Subscription percent circle

### LINKED ISSUE #208: Add % Circle to Member List
Subscribers "graduate" to alumni status after 2 years (730 days - we'll ignore leap years).

Add a member-visible % circle like this one:
https://codepen.io/VikasSharma/pen/oGLBEd

To this page:
/User

The page already lists the days of membership they have, based on their paid subscription. However, the % should only count days that have actually happened, not future paid-for days (which is what the subscription includes).

The % should be "past subscribed days" / 730. If the number is > 1 then the whole circle should be hidden and replaced with "Alumni".

ilyanaDev:
fixes #208 

Also adds alumni progress bar from MyProfile to member details

### DIFF
--- a/src/DevBetterWeb.Core/Interfaces/IMemberSubscriptionPeriodCalculationsService.cs
+++ b/src/DevBetterWeb.Core/Interfaces/IMemberSubscriptionPeriodCalculationsService.cs
@@ -10,5 +10,6 @@ namespace DevBetterWeb.Core.Interfaces
     DateTime GetCurrentSubscriptionEndDate(Member member);
     DateTime GetGraduationDate(Member member);
     int GetPercentageProgressToAlumniStatus(Member member);
+    int GetPercentageProgressToAlumniStatus(int memberSubscribedDays);
   }
 }
--- a/src/DevBetterWeb.Web/Pages/User/Details.cshtml.cs
+++ b/src/DevBetterWeb.Web/Pages/User/Details.cshtml.cs
@@ -16,7 +16,7 @@ namespace DevBetterWeb.Web.Pages.User
     private readonly IRepository<Member> _memberRepository;
     private readonly IMemberSubscriptionPeriodCalculationsService _memberSubscriptionPeriodCalculationsService;
 
-    public MemberSubscriptionPercentViewModel PercentModel { get; set; }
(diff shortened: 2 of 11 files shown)


src/DevBetterWeb.Core/Interfaces/IMemberSubscriptionPeriodCalculationsService.cs
```diff
@@ -10,5 +10,6 @@ public interface IMemberSubscriptionPeriodCalculationsService
     DateTime GetCurrentSubscriptionEndDate(Member member);
     DateTime GetGraduationDate(Member member);
     int GetPercentageProgressToAlumniStatus(Member member);
+    int GetPercentageProgressToAlumniStatus(int memberSubscribedDays);
```
ardalis, line 13: There's got to be an abstraction/data type you could use to return subscription details that doesn't involve so many primitive Get operations on this service. Unless there's a huge performance hit involved in any of these questions, just make a `SubscriptionDetails GetSubscriptionDetails(Member member)` call and then on that type add `CurrentSubscriptionEndDate` `GraduationDate` `PercentProgressToAlumniStatus` for instance.

src/DevBetterWeb.Web/Pages/User/Details.cshtml.cs
```diff
@@ -7,40 +7,50 @@
...
+    }
+
+    public async Task OnGet(string userId)
+    {
+      // I don't think we need this - SAS
+      //var user = await _userManager.FindByIdAsync(id);
+      //if (user == null)
+      //{
+      //    BadRequest();
+      //}
```
ardalis, line 35: Delete if you concur it's not needed.

ardalis (changes requested):

src/DevBetterWeb.Core/Interfaces/IMemberSubscriptionPeriodCalculationsService.cs
ilyanaDev: refactored out `GetPercentageProgressToAlumniStatus(int memberSubscribedDays);`
Better? 
This class _is_ supposed to be an abstraction to return subscription details.

PR #284, opened 2021-05-11
Cancellations report

### LINKED ISSUE #180: generate cancellations report for admins
generate and format a report for admins listing all cancellations by date for a given date range

ilyanaDev:
fixes #180 

Yes, there should be a partial for the various admin report pages, but I couldn't get that to work with the data binding for the dates, so I decided the small amount of tech debt from duplicating those pages was an acceptable trade-off to hitting my head against the wall trying to get the partial to work.

### DIFF
--- a/src/DevBetterWeb.Core/Specs/BillingActivitiesByDateTimeRangeAndCancelledVerbSpec.cs
+++ b/src/DevBetterWeb.Core/Specs/BillingActivitiesByDateTimeRangeAndCancelledVerbSpec.cs
@@ -10,8 +10,8 @@ namespace DevBetterWeb.Core.Specs
     public BillingActivitiesByDateTimeRangeAndCancelledVerbSpec(DateTimeRange range)
     {
       Query.Where(activity => activity.Details.ActionVerbPastTense == Enums.BillingActivityVerb.Cancelled)
-        .Where(activity => activity.Details.Date > range.StartDate)
-        .Where(activity => activity.Details.Date < range.EndDate)
+        .Where(activity => activity.Details.Date >= range.StartDate)
+        .Where(activity => activity.Details.Date <= range.EndDate)
         .OrderBy(activity => activity.Details.Date);
     }
   }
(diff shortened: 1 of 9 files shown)


src/DevBetterWeb.Core/Specs/BillingActivitiesByDateTimeRangeAndCancelledVerbSpec.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace DevBetterWeb.Core.Specs
+{
+  public class BillingActivitiesByDateTimeRangeAndCancelledVerbSpec : Specification<BillingActivity>
+  {
+    public BillingActivitiesByDateTimeRangeAndCancelledVerbSpec(DateTimeRange range)
+    {
+      Query.Where(activity => activity.Details.ActionVerbPastTense == Enums.BillingActivityVerb.Cancelled)
+        .Where(activity => activity.Details.Date > range.StartDate)
+        .Where(activity => activity.Details.Date < range.EndDate)
```
ardalis, line 14: Be sure these dates work as expected and there are no off by one errors. Should either of these be <= or >= ?

src/DevBetterWeb.Web/Pages/Admin/UserReports/Cancellations.cshtml.cs
```diff
@@ -0,0 +1,72 @@
...
+using Microsoft.AspNetCore.Mvc.RazorPages;
+
+namespace DevBetterWeb.Web.Pages.Admin.UserReports
+{
+  [Authorize(Roles = AuthConstants.Roles.ADMINISTRATORS)]
+
+  public class CancellationsModel : PageModel
+  {
+
+    private static readonly DateTimeRange defaultRange = new DateTimeRange(DateTime.Today.AddDays(-29), DateTime.Today.AddDays(-1));
```
ardalis, line 20: Understand this will show results through end of day-before-yesterday, not through yesterday, because of midnight default on datetimes. Is this what you want here?

src/DevBetterWeb.Web/Pages/Admin/UserReports/ReportsDatesModel.cs
```diff
@@ -0,0 +1,21 @@
...
+
+    public ReportsDatesModel()
+    {
+      // for model binding
+    }
+
+    public ReportsDatesModel(DateTime startDate, DateTime endDate)
+    {
+      StartDate = startDate;
+      EndDate = endDate;
```
ardalis, line 18: If you want end dates to be inclusive of all times on that date, you could add logic here to make enddate be date 23:59:59 on that date instead of 00:00:00. Then you wouldn't need to change things elsewhere. To make that clear you could rename this to something like InclusiveEndDate for instance.

ardalis (approved):

Issue #175, opened 2021-01-07
Allow Setting User Email Confirmation on Edit User Screen


ardalis:
On this page:
https://devbetter.com/Admin/User?userId={id}

Admins should be able to set the AspNetUsers.EmailConfirmed field to true or false.

patrykr5:
@ardalis please assign me to this

ShadyNagy:
It is already done.

PR #291, opened 2021-05-16
Added change email confirmation by admin and discord log for that

patrykr5:
Added feature #175


ardalis (approved):

ardalis:
Thanks! Some tests would be nice but I'll test it myself in a bit and maybe write some.

patrykr5:
@ardalis testing this feature is difficult and time-consuming as i'm using UserManager<ApplicationUser> (from .net identity). When i have some time i'll try to write something.

PR #309, opened 2021-05-26
New Member Subscription Logic

ilyanaDev:
Should now handle new Stripe subscriptions from current DevBetter members


src/DevBetterWeb.Core/Services/NewMemberService.cs
```diff
@@ -125,17 +127,6 @@ private Task AddUserToMemberRoleAsync(string userId)
...
-    {
-      var subscription = new Subscription();
-      subscription.MemberId = memberId;
-      subscription.Dates = subscriptionDateTimeRange;
-
-      var member = await _memberRepository.GetByIdAsync(memberId);
-      if (member is null) throw new MemberNotFoundException(memberId);
-      member.AddSubscription(subscription);
-    }
-
```
ardalis, line 138: I see this is the original code the service is replacing. I'm not sure what code calls this, but again if you shrink down subscription (use a constructor) and get the member elsewhere if needed, this code should shrink to almost nothing.

src/DevBetterWeb.Infrastructure/Services/UserLookupService.cs
```diff
@@ -22,5 +22,31 @@ public async Task<string> FindUserIdByEmailAsync(string email)
...
+
+      bool userIsMember = false;
+
+      foreach (var role in roles)
+      {
+        if (role.Equals(memberRoleName))
+        {
+          userIsMember = true;
+        }
+      }
```
ardalis, line 47: You could probably use some LINQ here. Something like:

```
return roles.Any(role => role.Equals(Constants.MEMBER_ROLE_NAME));
```

src/DevBetterWeb.Infrastructure/Services/WebhookHandlerService.cs
```diff
@@ -87,29 +98,77 @@ public async Task HandleNewCustomerSubscriptionAsync(string json)
...
 
-        var subscriptionPlanName = _paymentHandlerSubscription.GetAssociatedProductName(subscriptionId);
-        var billingPeriod = _paymentHandlerSubscription.GetBillingPeriod(subscriptionId);
-        await _memberAddBillingActivityService.AddSubscriptionCreationBillingActivity(email, paymentAmount, subscriptionPlanName, billingPeriod);
+      if (userIsMember)
+      {
+        return true;
       }
+
+      return false;
```
ardalis, line 154: it's hard to see in the code here but if the current code says this:

```
if (userIsMember)
{
  return true;
}
return false;
```
That's the same as this:

```
return userIsMember;
```

ardalis: And if that's harder to read for you, what about:

```
return userIsMember == true;
```

Which is also equivalent.

src/DevBetterWeb.Infrastructure/Services/WebhookHandlerService.cs
```diff
@@ -87,29 +98,77 @@ public async Task HandleNewCustomerSubscriptionAsync(string json)
...
+    {
+      var customerId = _paymentHandlerSubscription.GetCustomerId(subscriptionId);
+      var email = _paymentHandlerCustomer.GetCustomerEmail(customerId);
+
+      await AddNewSubscriberBillingActivity(subscriptionId, email, paymentAmount);
+
+      var subscriptionDateTimeRange = _paymentHandlerSubscription.GetDateTimeRange(subscriptionId);
+
+      var userId = await _userLookupService.FindUserIdByEmailAsync(email);
+      var spec = new MemberByUserIdSpec(userId);
```
ardalis, line 168: Is there no `MemberByEmailSpec`? If not, would it make sense to try and add one so you don't need the double lookup here to get a userId by email and then a member by userid?

ardalis (changes requested):

src/DevBetterWeb.Core/Services/NewMemberService.cs
```diff
@@ -101,7 +103,7 @@ public async Task<Member> MemberSetupAsync(string userId, string firstName, stri
 
       var subscriptionDateTimeRange = _paymentHandlerSubscription.GetDateTimeRange(subscriptionId);
 
-      await CreateSubscriptionForMemberAsync(member.Id, subscriptionDateTimeRange);
+      await _memberSubscriptionCreationService.CreateSubscriptionForMemberAsync(member.Id, subscriptionDateTimeRange);
 
       // Member has now been created and set up from the invite used. Invite should now be deactivated
       invite.Deactivate();
```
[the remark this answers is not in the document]
ilyanaDev: The first line of this method creates the member via the member registration service, which persists it (I think). And member.AddSubscription, called by the member subscription factory, should persist that

src/DevBetterWeb.Infrastructure/Services/WebhookHandlerService.cs
ilyanaDev: The problem is, we don't store email on member, so conceptually I feel like it's weird to do that? Even though it'll always be 1:1, so I guess it makes sense. I think this'd start to get complicated because you'd have the user lookup service involved. But perhaps adding a method/service somewhere to do this would probably be the way to go. It's certainly a useful functionality. Since I think it'd be somewhat complicated, I'll add an issue.

src/DevBetterWeb.Core/Services/MemberSubscriptionCreationService.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    public async Task CreateSubscriptionForMemberAsync(int memberId, DateTimeRange subscriptionDateTimeRange)
+    {
+      var subscription = new Subscription();
+      subscription.MemberId = memberId;
+      subscription.Dates = subscriptionDateTimeRange;
+
+      var member = await _memberRepository.GetByIdAsync(memberId);
+      if (member is null) throw new MemberNotFoundException(memberId);
+      member.AddSubscription(subscription);
```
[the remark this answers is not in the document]
ilyanaDev: That method is moved to Member.AddSubscription now. 

And yes, I'd agree that Subscription's constructor could use some refactoring, but I'll leave that for another time so that these other refactors can be finished.

ardalis (approved):

PR #337, opened 2021-06-14
subscription renamed to memberSubscription

ilyanaDev:


ardalis:
Merged. Need to fix prod schema now...

PR #342, opened 2021-06-21
Daily Check 

ilyanaDev:


src/DevBetterWeb.Core/Events/DailyCheckEvent.cs
```diff
@@ -0,0 +1,9 @@
+﻿using DevBetterWeb.Core.SharedKernel;
+
+namespace DevBetterWeb.Core.Events
+{
+  public class DailyCheckEvent : BaseDomainEvent
```
ardalis, line 5: Event names should be past tense. What happened? DailyCheckCompleted?

src/DevBetterWeb.Core/Specs/DailyCheckByDateSpec.cs
```diff
@@ -0,0 +1,15 @@
...
+using DevBetterWeb.Core.Entities;
+using System;
+
+namespace DevBetterWeb.Core.Specs
+{
+  public class DailyCheckByDateSpec : Specification<DailyCheck>, ISingleResultSpecification
+  {
+    public DailyCheckByDateSpec(DateTime date)
+    {
+      Query.Where(dc => dc.Date.Date.Equals(date.Date))
```
ardalis, line 11: Just checking, do you need `.dc.Date.Date` here?

src/DevBetterWeb.Core/Events/DailyCheckEvent.cs
ilyanaDev: well... it's not that the daily check was completed, it's an event saying "hey do the daily check." I'm not sure what a better name would be in that sense. Do you have any ideas?

src/DevBetterWeb.Core/Specs/DailyCheckByDateSpec.cs
ilyanaDev: yes. the first .date is the property, the second .date tells it to ignore whatever the time is on the datetime

src/DevBetterWeb.Core/Events/DailyCheckEvent.cs
ardalis: "Initiated" "Started"

src/DevBetterWeb.Web/Services/DailyCheckService.cs
```diff
@@ -0,0 +1,72 @@
...
+        return false;
+      }
+
+      return true;
+    }
+
+    private async void RaiseDailyCheckEvent()
+    {
+      await _dispatcher.Dispatch(new DailyCheckEvent());
+      _logger.LogInformation("Daily Check Event Raised");
```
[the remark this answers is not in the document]
ilyanaDev: sure

src/DevBetterWeb.Core/Specs/DailyCheckByDateSpec.cs
ardalis: Give the property a more descriptive name. I think BaseDomainEvent already has a date on it, but since this is the one you want to record with the "DailyCheck" perhaps it should be "CheckInitiatedDateTime"?

It would probably be good to store the full date time value with the DailyCheck, which I think you're doing. In that case, it's usually more clear to use `DateTime` in the property name rather than `Date` just to make that obvious.

PR #344, opened 2021-06-21
Daily check for graduation

ilyanaDev:
devbetter should now automate moving people to alumni status on the website, as well as sending them and admins emails notifying that they've graduated

admins will still need to update discord roles and at this point still deal with payment stuff

### DIFF
--- a/src/DevBetterWeb.Infrastructure/Handlers/DailyCheckInitiatedEventHandler.cs
+++ b/src/DevBetterWeb.Infrastructure/Handlers/DailyCheckInitiatedEventHandler.cs
@@ -1,0 +1,30 @@
+﻿using DevBetterWeb.Core.Events;
+using DevBetterWeb.Core.Interfaces;
+using DevBetterWeb.Infrastructure.Services;
+using System.Threading.Tasks;
+
+namespace DevBetterWeb.Core.Handlers
+{
+  public class DailyCheckInitiatedEventHandler : IHandle<DailyCheckInitiatedEvent>
+  {
+    private readonly AdminUpdatesWebhook _webhook;
+
+    public DailyCheckInitiatedEventHandler(AdminUpdatesWebhook webhook)
+    {
+      _webhook = webhook;
+    }
+
+    public async Task Handle(DailyCheckInitiatedEvent domainEvent)
+    {
+      
+      // Add real stuff
+
+      // check if people need upgraded to alumni
+
+      // check if people need to be pinged about new member link
+
+      _webhook.Content = "Daily Check Event Completed";
+      await _webhook.Send();
+    }
+  }
+}
--- a/src/DevBetterWeb.Infrastructure/Handlers/DailyCheckInitiatedEventHandler.cs
+++ b/src/DevBetterWeb.Infrastructure/Handlers/DailyCheckInitiatedEventHandler.cs
@@ -21,15 +21,23 @@ namespace DevBetterWeb.Core.Handlers
     public async Task Handle(DailyCheckInitiatedEvent domainEvent)
     {
       AppendOnlyStringList messages = new();
-      
-      // Add real stuff
+
+      // real stuff
+
+      // check if admins need to be reminded to avoid renewing near-alumnus's subscription
 
       // check if people need upgraded to alumni
+      await _alumniGraduationService.GraduateMembersIfNeeded(messages);
 
       // check if people need to be pinged about new member link
 
-      _webhook.Content = "Daily Check Event Completed";
-      await _webhook.Send();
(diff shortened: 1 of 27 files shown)


src/DevBetterWeb.Infrastructure/Handlers/DailyCheckInitiatedEventHandler.cs
```diff
@@ -1,30 +1,43 @@
...
 
-      _webhook.Content = "Daily Check Event Completed";
-      await _webhook.Send();
+      messages.Append("Daily Check Event Completed");
+
+      foreach (var message in messages)
+      {
+        _webhook.Content = message;
+        await _webhook.Send();
+      }
```
ardalis, line 40: Nice use of loop

src/DevBetterWeb.Infrastructure/Services/AlumniGraduationService.cs
```diff
@@ -0,0 +1,102 @@
...
+using DevBetterWeb.Core.Interfaces;
+using DevBetterWeb.Infrastructure.Identity.Data;
+using Microsoft.AspNetCore.Identity;
+
+namespace DevBetterWeb.Infrastructure.Services
+{
+  public class AlumniGraduationService : IAlumniGraduationService
+  {
+    private const int DAYS_IN_TWO_YEARS = 365 * 2;
+
```
ardalis, line 15: I like using this to make the final number clear.

src/DevBetterWeb.Infrastructure/Services/AlumniGraduationService.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+      _userLookupService = userLookupService;
+      _repository = repository;
+      _graduationCommunicationsService = graduationCommunicationsService;
+      _userManager = userManager;
+    }
+
+    public async Task GraduateMembersIfNeeded(AppendOnlyStringList messages)
+    {
+      var members = await _repository.ListAsync();
```
ardalis, line 34: Do you need all members? Would just a list of non-alumni members be more efficient?

src/DevBetterWeb.Infrastructure/Services/GraduationCommunicationsService.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+      _emailService = emailService;
+      _userManager = userManager;
+    }
+
+    public async Task SendGraduationCommunications(Member member)
+    {
+      var memberEmail = "";
+      var memberSubject = "Congratulations from DevBetter!";
+      var memberText = "You're now an alumnus of DevBetter! Congratulations on your graduation!\nNow that you're an alumnus, you'll retain full access as long as you want it, but your subscription is free.";
```
ardalis, line 27: "as long as you want it (and you follow our code of conduct and other terms), ..."

tests/DevBetterWeb.Tests/Services/AlumniGraduationServiceTests/CheckIfAnyMemberGraduating.cs
```diff
@@ -0,0 +1,144 @@
...
+
+    private AlumniGraduationService service;
+
+    public CheckIfAnyMemberGraduating()
+    {
+      userLookupService = new Mock<IUserLookupService>();
+      repository = new Mock<IRepository<Member>>();
+      graduationCommunications = new Mock<IGraduationCommunicationsService>();
+      var store = new Mock<IUserStore<ApplicationUser>>();
+      userManager = new Mock<UserManager<ApplicationUser>>(store.Object, null, null, null, null, null, null, null, null);
```
ardalis, line 34: this big interface you have to mock is a code smell. Would it be possible to define a smaller interface that only had the method(s) you need from UserManager, and mock that instead? Then you'd have an adapter somewhere that implemented the small interface with Usermanager, but this code wouldn't care about that.

src/DevBetterWeb.Infrastructure/Services/AlumniGraduationService.cs
ilyanaDev: :)

src/DevBetterWeb.Infrastructure/Services/AlumniGraduationService.cs
ilyanaDev: Yes, that would be more efficient. But I don't think I can really define that in a spec because I have to go from member to its userId to the corresponding user to that user's roles. Which would be a lot of lambda expressions that I don't think a spec can handle. If I'm wrong on that I could change it to that though

PR #354, opened 2021-07-01
Update with test url details for discord webhooks

samuelwine:


ardalis (approved):

ardalis:
@samuelwine Good addition, thanks. Also you could wire up a "mock" service for the discord notification service, currently implemented in BaseWebHook. Looking at the code, though, there's not a single interface involved, so it wouldn't be a simple task. I'll work toward making that easier.

PR #371, opened 2021-07-14
Reminder emails for new signups

ilyanaDev:
DailyCheckPingService added to remind new users (who have paid but not set up their account) to set up their account if they have not within 2 days of paying. After 4 days (4 days after user paid, 2 days after user ping), admins get a ping to remind the user directly.

This shouldn't do anything until the switch to stripe.

It will require a db update, as I added some new properties to Invitation to allow this functionality.

### DIFF
--- a/src/DevBetterWeb.Core/Entities/Invitation.cs
+++ b/src/DevBetterWeb.Core/Entities/Invitation.cs
@@ -11,8 +11,8 @@ namespace DevBetterWeb.Core.Entities
     public string PaymentHandlerSubscriptionId { get; private set; }
     public bool Active { get; private set; } = true;
     public DateTime DateCreated { get; private set; }
-    public DateTime? DateOfUserPing { get; private set; }
-    public DateTime? DateOfLastAdminPing { get; private set; }
+    public DateTime DateOfUserPing { get; private set; } = DateTime.MinValue;
+    public DateTime DateOfLastAdminPing { get; private set; } = DateTime.MinValue;
 
     public Invitation(string email, string inviteCode, string paymentHandlerSubscriptionId)
     {
@@ -20,6 +20,17 @@ namespace DevBetterWeb.Core.Entities
       InviteCode = inviteCode;
       PaymentHandlerSubscriptionId = paymentHandlerSubscriptionId;
       DateCreated = DateTime.Today;
+    }
+
+    // for testing
+    public Invitation(string email, string inviteCode, string paymentHandlerSubscriptionId, DateTime dateCreated, DateTime dateOfUserPing, DateTime dateOfLastAdminPing)
+    {
+      Email = email;
+      InviteCode = inviteCode;
+      PaymentHandlerSubscriptionId = paymentHandlerSubscriptionId;
+      DateCreated = dateCreated;
+      DateOfUserPing = dateOfUserPing;
+      DateOfLastAdminPing = dateOfLastAdminPing;
     }
 
     private Invitation()
@@ -32,6 +43,15 @@ namespace DevBetterWeb.Core.Entities
     public void Deactivate()
     {
       Active = false;
+    }
+
+    public void UpdateUserPingDate()
+    {
+      DateOfUserPing = DateTime.Today;
+    }
+    public void UpdateAdminPingDate()
+    {
+      DateOfLastAdminPing = DateTime.Today;
     }
   }
 }
--- a/src/DevBetterWeb.Core/Entities/Invitation.cs
+++ b/src/DevBetterWeb.Core/Entities/Invitation.cs
@@ -1,4 +1,5 @@
-﻿using DevBetterWeb.Core.Interfaces;
+﻿using System;
+using DevBetterWeb.Core.Interfaces;
 using DevBetterWeb.Core.SharedKernel;
 
 namespace DevBetterWeb.Core.Entities
@@ -9,12 +10,16 @@ namespace DevBetterWeb.Core.Entities
     public string InviteCode { get; private set; }
     public string PaymentHandlerSubscriptionId { get; private set; }
     public bool Active { get; private set; } = true;
+    public DateTime DateCreated { get; private set; }
+    public DateTime? DateOfUserPing { get; private set; }
+    public DateTime? DateOfLastAdminPing { get; private set; }
 
     public Invitation(string email, string inviteCode, string paymentHandlerSubscriptionId)
     {
       Email = email;
       InviteCode = inviteCode;
       PaymentHandlerSubscriptionId = paymentHandlerSubscriptionId;
+      DateCreated = DateTime.Today;
     }
 
     private Invitation()
--- a/src/DevBetterWeb.Core/Entities/Invitation.cs
+++ b/src/DevBetterWeb.Core/Entities/Invitation.cs
@@ -49,6 +49,7 @@ namespace DevBetterWeb.Core.Entities
     {
       DateOfUserPing = DateTime.Today;
     }
+
     public void UpdateAdminPingDate()
     {
       DateOfLastAdminPing = DateTime.Today;
--- a/src/DevBetterWeb.Core/Interfaces/ApplicationUser.cs
+++ b/src/DevBetterWeb.Core/Interfaces/ApplicationUser.cs
@@ -1,0 +1,6 @@
+﻿namespace DevBetterWeb.Core.Interfaces
+{
+  internal class ApplicationUser
+  {
+  }
+}
--- a/src/DevBetterWeb.Core/Interfaces/ApplicationUser.cs
+++ b/src/DevBetterWeb.Core/Interfaces/ApplicationUser.cs
@@ -1,6 +1,0 @@
-﻿namespace DevBetterWeb.Core.Interfaces
-{
-  internal class ApplicationUser
-  {
-  }
-}
--- a/src/DevBetterWeb.Core/Interfaces/IDailyCheckPingService.cs
+++ b/src/DevBetterWeb.Core/Interfaces/IDailyCheckPingService.cs
@@ -7,7 +7,5 @@ namespace DevBetterWeb.Core.Interfaces
   public interface IDailyCheckPingService
   {
     Task SendPingIfNeeded(AppendOnlyStringList messages);
-    List<Invitation> CheckIfAnyActiveInvitationsRequireUserPing(List<Invitation> invitations);
-    List<Invitation> CheckIfAnyActiveInvitationsRequireAdminPing(List<Invitation> invitations);
   }
 }
--- a/src/DevBetterWeb.Core/Interfaces/IDailyCheckPingService.cs
+++ b/src/DevBetterWeb.Core/Interfaces/IDailyCheckPingService.cs
@@ -1,0 +1,13 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using DevBetterWeb.Core.Entities;
+
+namespace DevBetterWeb.Core.Interfaces
+{
+  public interface IDailyCheckPingService
+  {
+    Task SendPingIfNeeded(AppendOnlyStringList messages);
+    List<Invitation> CheckIfAnyActiveInvitationsRequireUserPing(List<Invitation> invitations);
+    List<Invitation> CheckIfAnyActiveInvitationsRequireAdminPing(List<Invitation> invitations);
+  }
+}
--- a/src/DevBetterWeb.Core/Specs/ActiveInvitationsSpec.cs
+++ b/src/DevBetterWeb.Core/Specs/ActiveInvitationsSpec.cs
@@ -1,0 +1,13 @@
+﻿using Ardalis.Specification;
+using DevBetterWeb.Core.Entities;
+
+namespace DevBetterWeb.Core.Specs
+{
+  public class ActiveInvitationsSpec : Specification<Invitation>
+  {
+    public ActiveInvitationsSpec()
+    {
+      Query.Where(invite => invite.Active);
+    }
+  }
+}
--- a/src/DevBetterWeb.Core/Specs/InactiveInvitationByEmailSpec.cs
+++ b/src/DevBetterWeb.Core/Specs/InactiveInvitationByEmailSpec.cs
@@ -3,7 +3,7 @@ using DevBetterWeb.Core.Entities;
 
 namespace DevBetterWeb.Core.Specs
 {
-  public partial class InactiveInvitationByEmailSpec : Specification<Invitation>, ISingleResultSpecification
+  public class InactiveInvitationByEmailSpec : Specification<Invitation>, ISingleResultSpecification
   {
     public InactiveInvitationByEmailSpec(string email)
     {
--- a/src/DevBetterWeb.Core/Specs/InactiveInvitationByEmailSpec.cs
+++ b/src/DevBetterWeb.Core/Specs/InactiveInvitationByEmailSpec.cs
@@ -3,7 +3,7 @@ using DevBetterWeb.Core.Entities;
 
 namespace DevBetterWeb.Core.Specs
 {
-  public class InactiveInvitationByEmailSpec : Specification<Invitation>, ISingleResultSpecification
+  public partial class InactiveInvitationByEmailSpec : Specification<Invitation>, ISingleResultSpecification
   {
     public InactiveInvitationByEmailSpec(string email)
     {
--- a/src/DevBetterWeb.Infrastructure/Handlers/DailyCheckInitiatedEventHandler.cs
+++ b/src/DevBetterWeb.Infrastructure/Handlers/DailyCheckInitiatedEventHandler.cs
@@ -42,20 +42,35 @@ namespace DevBetterWeb.Core.Handlers
 
       messages.Append("Daily Check Event Completed");
 
-      var spec = new DailyCheckByDateSpec(DateTime.Today);
+      await SendMessagesToDiscord(messages);
+      await StoreMessagesInTasksCompleted(messages);
+    }
 
-      var todaysDailyCheck = await _repository.GetBySpecAsync(spec);
-
+    private async Task SendMessagesToDiscord(AppendOnlyStringList messages)
+    {
       foreach (var message in messages)
       {
         _webhook.Content = message;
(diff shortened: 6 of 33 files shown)


src/DevBetterWeb.Core/Entities/Invitation.cs
```diff
@@ -9,12 +10,27 @@ public class Invitation : BaseEntity, IAggregateRoot
...
+    public DateTime DateCreated { get; private set; }
+    public DateTime DateOfUserPing { get; private set; } = DateTime.MinValue;
+    public DateTime DateOfLastAdminPing { get; private set; } = DateTime.MinValue;
 
     public Invitation(string email, string inviteCode, string paymentHandlerSubscriptionId)
     {
       Email = email;
       InviteCode = inviteCode;
       PaymentHandlerSubscriptionId = paymentHandlerSubscriptionId;
+      DateCreated = DateTime.Today;
```
ardalis, line 22: You can do this on line 13 when you define the property.

src/DevBetterWeb.Core/Entities/Invitation.cs
```diff
@@ -9,12 +10,27 @@ public class Invitation : BaseEntity, IAggregateRoot
...
+    }
+
+    // for testing
```
ardalis, line 25: Why do you need a separate constructor for testing? That's usually a code smell. Can you not configure the state of the type using its regular constructor plus properties?

src/DevBetterWeb.Core/Entities/Invitation.cs
```diff
@@ -9,12 +10,27 @@ public class Invitation : BaseEntity, IAggregateRoot
...
+    {
+      Email = email;
+      InviteCode = inviteCode;
+      PaymentHandlerSubscriptionId = paymentHandlerSubscriptionId;
+      DateCreated = dateCreated;
+      DateOfUserPing = dateOfUserPing;
+      DateOfLastAdminPing = dateOfLastAdminPing;
     }
 
     private Invitation()
```
ardalis, line 36: I think this could be empty (or perhaps eliminated) if all of these sets were done where the properties are defined.

It's generally preferable to set property defaults inline since it's easier to see whether you forgot to do so with any of them without the vertical separation between defining them and initializing them in the constructor.

src/DevBetterWeb.Core/Entities/Invitation.cs
```diff
@@ -28,5 +44,14 @@ public void Deactivate()
     {
       Active = false;
     }
+
+    public void UpdateUserPingDate()
```
ardalis, line 48: If you needed it, you could pass in the datetime variable but have it default to DateTime.Today:

`public void UpdateUserPingDate(DateTime theDate = DateTime.Today)`

Something close to that should work.

src/DevBetterWeb.Core/Interfaces/ApplicationUser.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace DevBetterWeb.Core.Interfaces
+{
+  internal class ApplicationUser
```
ardalis, line 3: What is this doing in the interfaces namespace?

src/DevBetterWeb.Core/Interfaces/IDailyCheckPingService.cs
```diff
@@ -0,0 +1,13 @@
...
+using DevBetterWeb.Core.Entities;
+
+namespace DevBetterWeb.Core.Interfaces
+{
+  public interface IDailyCheckPingService
+  {
+    Task SendPingIfNeeded(AppendOnlyStringList messages);
+    List<Invitation> CheckIfAnyActiveInvitationsRequireUserPing(List<Invitation> invitations);
+    List<Invitation> CheckIfAnyActiveInvitationsRequireAdminPing(List<Invitation> invitations);
+  }
```
ardalis, line 12: This is fine for now but has some obvious SRP and OCP violation issues.

A more extensible design would be to have a generic service that performs 0 to N checks, and another interface that defines a single check. The generic service would get an IEnumerable<CheckService> and loop over them, performing the checks. Then you could keep adding checks without having to add methods to an interface. Make sense?

src/DevBetterWeb.Core/Specs/ActiveInvitationsSpec.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Ardalis.Specification;
+using DevBetterWeb.Core.Entities;
+
+namespace DevBetterWeb.Core.Specs
+{
+  public class ActiveInvitationsSpec : Specification<Invitation>
```
ardalis, line 6: Looks good!

src/DevBetterWeb.Core/Specs/InactiveInvitationByEmailSpec.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace DevBetterWeb.Core.Specs
 {
-  public class InactiveInvitationByEmailSpec : Specification<Invitation>, ISingleResultSpecification
+  public partial class InactiveInvitationByEmailSpec : Specification<Invitation>, ISingleResultSpecification
```
ardalis, line 6: Why is this partial?

src/DevBetterWeb.Infrastructure/Handlers/DailyCheckInitiatedEventHandler.cs
```diff
@@ -29,13 +38,24 @@ public async Task Handle(DailyCheckInitiatedEvent domainEvent)
...
       foreach (var message in messages)
       {
         _webhook.Content = message;
         await _webhook.Send();
+
+        if(todaysDailyCheck != null && !message.Equals("Daily Check Event Completed"))
+        {
+          if (todaysDailyCheck.TasksCompleted == null) todaysDailyCheck.TasksCompleted = "";
+          todaysDailyCheck.TasksCompleted += message;
+        }
```
ardalis, line 58: This seems overly complicated and reliant upon magic strings, which you're duplicating on lines 43 and 54. Is there a way to simplify this logic?

src/DevBetterWeb.Infrastructure/Services/DailyCheckPingService.cs
```diff
@@ -0,0 +1,150 @@
...
+      {
+        bool inviteRequiresAdminPing = true;
+
+        if (invite.DateOfUserPing.Equals(DateTime.MinValue)) inviteRequiresAdminPing = false;
+        if (invite.DateOfLastAdminPing.Equals(DateTime.Today)) inviteRequiresAdminPing = false;
+        if (DateTime.Today - invite.DateCreated < new TimeSpan(4, 0, 0, 0)) inviteRequiresAdminPing = false;
+        if (DateTime.Today - invite.DateOfUserPing < new TimeSpan(2, 0, 0, 0)) inviteRequiresAdminPing = false;
+        if (!invite.Active) inviteRequiresAdminPing = false;
+
+        if (inviteRequiresAdminPing) invitationsRequiringAdminPing.Add(invite);
```
ardalis, line 78: Make `inviteRequiresAdminPing` a method and move all the other if statements except this one into that method (which should return bool). This will make the foreach loop much smaller and easier to follow.

src/DevBetterWeb.Infrastructure/Services/DailyCheckPingService.cs
```diff
@@ -0,0 +1,150 @@
...
+      var invitationsRequiringAdminPing = new List<Invitation>();
+
+      foreach(var invite in invitations)
+      {
+        bool inviteRequiresAdminPing = true;
+
+        if (invite.DateOfUserPing.Equals(DateTime.MinValue)) inviteRequiresAdminPing = false;
+        if (invite.DateOfLastAdminPing.Equals(DateTime.Today)) inviteRequiresAdminPing = false;
+        if (DateTime.Today - invite.DateCreated < new TimeSpan(4, 0, 0, 0)) inviteRequiresAdminPing = false;
+        if (DateTime.Today - invite.DateOfUserPing < new TimeSpan(2, 0, 0, 0)) inviteRequiresAdminPing = false;
```
ardalis, line 75: If you have magic numbers (4, 2) it's best to name them somewhere. If they're only used here, they should be named here. So you might say something like:

TimeSpan remind4DaysAfterCreation = new TimeSpan(4,0,0,0);
TimeSpan remind2DaysAfterLastPing = new TimeSpan(2,0,0,0);

Then your if statements would use these variables, not the magic timespan values directly.

src/DevBetterWeb.Infrastructure/Services/DailyCheckPingService.cs
```diff
@@ -0,0 +1,150 @@
...
+      return invitationsRequiringAdminPing;
+    }
+
+    public List<Invitation> CheckIfAnyActiveInvitationsRequireUserPing(List<Invitation> invitations)
+    {
+      var invitationsRequiringUserPing = new List<Invitation>();
+
+      foreach(var invite in invitations)
+      {
+        if((DateTime.Today - invite.DateCreated) >= new TimeSpan(2,0,0,0) && invite.DateOfUserPing.Equals(DateTime.MinValue))
```
ardalis, line 90: Magic TimeSpan

src/DevBetterWeb.Infrastructure/Services/DailyCheckPingService.cs
```diff
@@ -0,0 +1,150 @@
...
+      foreach (var invitation in invitations)
+      {
+        listOfEmails += $"{invitation.Email}\n";
+      }
+
+      var emailBody = $"Please remind these users to finish setting up their DevBetter accounts: {listOfEmails}";
+
+      var usersInAdminRole = await _userManager.GetUsersInRoleAsync(AuthConstants.Roles.ADMINISTRATORS);
+
+      var emailList = "";
```
ardalis, line 131: What's the difference between `listOfEmails` and `emailList`?

It seems like the latter is the list of sent emails while the former is the list of emails to send. Name them accordingly.

tests/DevBetterWeb.Tests/Services/DailyCheckPingServiceTests/CheckIfAnyActiveInvitationsRequireAdminPing.cs
```diff
@@ -0,0 +1,159 @@
...
+      _userManager = new Mock<UserManager<ApplicationUser>>(store.Object, null, null, null, null, null, null, null, null);
+      _emailService = new Mock<IEmailService>();
+      _repository = new Mock<IRepository<Invitation>>();
+      service = new DailyCheckPingService(_repository.Object, _emailService.Object, _userManager.Object);
+    }
+
+    [Fact]
+    public void ReturnsEmptyListGivenEmptyList()
+    {
+      var testlist = new List<Invitation>();
```
ardalis, line 33: This line appears 9 times in this file. It could probably be pulled up into a private field and initialized inline (or in the constructor if you prefer).

tests/DevBetterWeb.Tests/Services/DailyCheckPingServiceTests/CheckIfAnyActiveInvitationsRequireAdminPing.cs
```diff
@@ -0,0 +1,159 @@
...
+      testlist.Add(valid);
+      testlist.Add(invalid);
+
+      var list = service.CheckIfAnyActiveInvitationsRequireAdminPing(testlist);
+
+      Assert.Contains(valid, list);
+      Assert.DoesNotContain(invalid, list);
+    }
```
ardalis, line 133: Good!

tests/DevBetterWeb.Tests/Services/DailyCheckPingServiceTests/CheckIfAnyActiveInvitationsRequireUserPing.cs
```diff
@@ -0,0 +1,112 @@
...
+      _userManager = new Mock<UserManager<ApplicationUser>>(store.Object, null, null, null, null, null, null, null, null);
+      _emailService = new Mock<IEmailService>();
+      _repository = new Mock<IRepository<Invitation>>();
+      service = new DailyCheckPingService(_repository.Object, _emailService.Object, _userManager.Object);
+    }
+
+    [Fact]
+    public void ReturnsEmptyListGivenEmptyList()
+    {
+      var testlist = new List<Invitation>();
```
ardalis, line 36: Pull up to field.

ardalis (changes requested):

src/DevBetterWeb.Core/Entities/Invitation.cs
ilyanaDev: I needed to test logic based on when the various events (creation, user ping, admin ping) happened but those shouldn't really be able to be publically set. Especially creation, which (except in testing) should just happen in the constructor. Definitely a code smell but I'm not sure of a better way to do it as I was trying to avoid making these things have public setters

src/DevBetterWeb.Core/Entities/Invitation.cs
ilyanaDev: these aren't defaults, they're generally test values. Defaults _are_ set where the properties are defined

src/DevBetterWeb.Core/Entities/Invitation.cs
ilyanaDev: ah that's smart

ilyanaDev: though I'm not sure I want people to be able to mess with that

ilyanaDev: at least when I'm using the constructor no one can change the date after the invitation is created. Which is probably how it should be

src/DevBetterWeb.Core/Interfaces/ApplicationUser.cs
ilyanaDev: no idea. Probably accidentally added it when using user manager<applicationUser> somewhere

src/DevBetterWeb.Core/Interfaces/IDailyCheckPingService.cs
ilyanaDev: I see. Well, actually, the only reason these are public is so I can test them, because they're the main logic of the class. But I can have public methods on the implementation that aren't on the interface, right? Because they're called from SendPingIfNeeded

src/DevBetterWeb.Core/Specs/ActiveInvitationsSpec.cs
ilyanaDev: :)

src/DevBetterWeb.Core/Specs/InactiveInvitationByEmailSpec.cs
ilyanaDev: excellent question. Not sure. Probably copy-paste error or autocomplete error

src/DevBetterWeb.Infrastructure/Handlers/DailyCheckInitiatedEventHandler.cs
ilyanaDev: do you want to have to store "daily check event completed" in the db for every day, or can I omit that? That's basically all this is trying to accomplish. But I'll at least pull some of it into its own method so it's a bit more clear

src/DevBetterWeb.Core/Entities/Invitation.cs
ardalis: Something else I missed - don't set things to DateTime.MinValue as defaults or to mean "never happened". Instead, use a nullable DateTime (`DateTime?`) and default them to null. If that makes it complain about nullable reference types you can override the warning.

If creation always defaults to Today, that should be easily unit tested. There would only be rare edge cases where you ran the tests near midnight when a failure would occur. So no need to have a separate way to set that, typically. Instead you might have a way of abstracting time in your service, so that even if the invite is created Today, the service might be executing on Today.AddDays(5) which you could specify via an IDateTime mock that is injected into the service, for instance.

The other dates must have some way of being set so probably you could call whatever method sets them to put them into a valid state. Having a separate test constructor works but is usually a last resort.

src/DevBetterWeb.Core/Entities/Invitation.cs
ardalis: I was referencing the contents of the `private Invitation()` constructor, which just has things set to "" and such.

src/DevBetterWeb.Core/Entities/Invitation.cs
ardalis: That's only reasonable for the CreationDate; the other dates will change.

src/DevBetterWeb.Core/Interfaces/IDailyCheckPingService.cs
ardalis: Yes, you can certainly have public methods that aren't on the interface.

src/DevBetterWeb.Infrastructure/Services/DailyCheckPingService.cs
ilyanaDev: ok

src/DevBetterWeb.Infrastructure/Handlers/DailyCheckInitiatedEventHandler.cs
ardalis: no need to store that in the db every day, no. no need to keep/persist a log of every check - the discord history will do a good enough job of that.

tests/DevBetterWeb.Tests/Services/DailyCheckPingServiceTests/CheckIfAnyActiveInvitationsRequireAdminPing.cs
ilyanaDev: :)

ardalis:
We can merge this in when you're ready and then do another PR to cover outstanding refactorings like:

- Use builder pattern in tests
- Use reflection to set private properties rather than expose them as public properties or public constructor
- Upgrade messages from strings to a class with a (fixed) message and a `shouldPersist` boolean property

PR #373, opened 2021-07-14
tests of daily check ping service 

ilyanaDev:
refactored to use reflection and builder pattern

### DIFF
--- a/tests/DevBetterWeb.Tests/Services/DailyCheckPingServiceTests/InvitationBuilder.cs
+++ b/tests/DevBetterWeb.Tests/Services/DailyCheckPingServiceTests/InvitationBuilder.cs
@@ -3,37 +3,39 @@ using DevBetterWeb.Core.Entities;
 
 namespace DevBetterWeb.Tests.Services.DailyCheckPingServiceTests
 {
-  public static class InvitationBuilder
+  public class InvitationBuilder
   {
     public const string DEFAULT_EMAIL = "test@email.com";
     public const string DEFAULT_INVITE_CODE = "defaultInviteCode";
     public const string DEFAULT_SUBSCRIPTION_ID = "testSubscriptionId";
 
-    public static Invitation BuildDefaultInvitation()
+    private Invitation _invitation = new Invitation(DEFAULT_EMAIL, DEFAULT_INVITE_CODE, DEFAULT_SUBSCRIPTION_ID);
+
+    public InvitationBuilder WithDateCreatedGivenDaysAgo(int daysAgo)
     {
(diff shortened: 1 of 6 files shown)


tests/DevBetterWeb.Tests/Services/DailyCheckPingServiceTests/InvitationBuilder.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using DevBetterWeb.Core.Entities;
+
+namespace DevBetterWeb.Tests.Services.DailyCheckPingServiceTests
+{
+  public static class InvitationBuilder
```
ardalis, line 6: Builders typically shouldn't be static. If you run your tests in parallel, the shared state of static builders will cause random errors. Make it non-static, and then each test class that needs it should have a private field that is the builder that is instantiated when the test is created:

`private InvitationBuilder _invitationBuilder = new();`

Then use `_invitationBuilder` anywhere you need it in your tests.

tests/DevBetterWeb.Tests/Services/DailyCheckPingServiceTests/InvitationBuilder.cs
```diff
@@ -0,0 +1,39 @@
...
+  {
+    public const string DEFAULT_EMAIL = "test@email.com";
+    public const string DEFAULT_INVITE_CODE = "defaultInviteCode";
+    public const string DEFAULT_SUBSCRIPTION_ID = "testSubscriptionId";
+
```
ardalis, line 11: Good use of constants!

tests/DevBetterWeb.Tests/Services/DailyCheckPingServiceTests/ReflectionHelperExtensionMethods.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using DevBetterWeb.Core.Entities;
+using System.Reflection;
+
+namespace DevBetterWeb.Tests.Services.DailyCheckPingServiceTests
+{
+  public static class ReflectionHelperExtensionMethods
+  {
+    // from https://stackoverflow.com/a/1565766/13680266
+    public static void SetPrivateDateTimePropertyValue(this Invitation invitation, string propName, DateTime newValue)
```
ardalis, line 10: good attribution

ardalis (changes requested):

ardalis (approved):
This looks great!

Some things to think about blogging about:

- Setting up private state of classes under test. Different options and pros/cons.
- Using builders/static factory methods/etc for test data setup, and creating a pseudo-Builder using statics and extension methods

PR #379, opened 2021-07-20
Member subscription plan addition

ilyanaDev:
CRUD available to admins for MemberSubscriptionPlan

### DIFF
--- a/src/DevBetterWeb.Web/DevBetterWeb.Web.csproj
+++ b/src/DevBetterWeb.Web/DevBetterWeb.Web.csproj
@@ -34,7 +34,7 @@
 		<PackageReference Include="MediatR" Version="9.0.0" />
 		<PackageReference Include="MediatR.Extensions.Microsoft.DependencyInjection" Version="9.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation" Version="5.0.8" />
-		<PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="5.0.6" />
+		<PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="5.0.8" />
 		<PackageReference Include="Microsoft.Extensions.Logging.AzureAppServices" Version="5.0.8" />
 		<!-- added to fix error when upgrading to .net 5.0 -->
 		<PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="5.0.2" />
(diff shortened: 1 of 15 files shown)


ilyanaDev:
not sure why the build is failing. It works locally. I've tried re running it and that hasn't made a difference.

src/DevBetterWeb.Web/DevBetterWeb.Web.csproj
```diff
@@ -34,15 +34,15 @@
 		<PackageReference Include="MediatR" Version="9.0.0" />
 		<PackageReference Include="MediatR.Extensions.Microsoft.DependencyInjection" Version="9.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation" Version="5.0.8" />
-		<PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="5.0.8" />
+		<PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="5.0.6" />
```
ardalis, line 37: Why are we downgrading these?

src/DevBetterWeb.Web/Pages/Admin/Books/Create.cshtml.cs
```diff
@@ -1,13 +1,16 @@
...
 using DevBetterWeb.Core.Events;
 using DevBetterWeb.Core.Interfaces;
 using DevBetterWeb.Infrastructure.Data;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
 
 namespace DevBetterWeb.Web.Pages.Admin.Books
 {
+  [Authorize(Roles = AuthConstants.Roles.ADMINISTRATORS)]
```
ardalis, line 13: This is good to lock down, thanks.

src/DevBetterWeb.Web/Pages/Admin/SubscriptionPlans/Delete.cshtml.cs
```diff
@@ -0,0 +1,62 @@
...
+            if (id == null)
+            {
+                return NotFound();
+            }
+
+            MemberSubscriptionPlan = await _context.MemberSubscriptionPlan.FindAsync(id);
+
+            if (MemberSubscriptionPlan != null)
+            {
+                _context.MemberSubscriptionPlan.Remove(MemberSubscriptionPlan);
```
ardalis, line 55: Do we want to add any checks to this before allowing deletion? Maybe see if the plan is actively being used by a subscription? If we care about that, should we show it in the list of plans ("Used by N Active Subscriptions")?

src/DevBetterWeb.Web/Pages/Admin/SubscriptionPlans/Index.cshtml.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        public IList<MemberSubscriptionPlan>? MemberSubscriptionPlansList { get;set; }
+
+        public async Task OnGetAsync()
+        {
+            MemberSubscriptionPlansList = await _context.MemberSubscriptionPlan!.AsQueryable().ToListAsync();
+        }
+    }
+}
```
ardalis, line 32: See if you can fix some of the warnings in the GitHub build. They seem to mostly be possible null value dereferences and async methods missing await keywords. Otherwise looks good.

ardalis (approved):

src/DevBetterWeb.Web/DevBetterWeb.Web.csproj
ilyanaDev: not sure. Don't think I did that on purpose. But I fixed it

PR #380, opened 2021-07-21
daily check for upcoming graduation

ilyanaDev:
daily check added to remind admins to make sure upcoming alumni do not have subscription renewed
eventually this should be replaced with an automatic call to stripe to cancel the subscription in question

### DIFF
--- a/src/DevBetterWeb.Infrastructure/Services/DailyCheckPingService.cs
+++ b/src/DevBetterWeb.Infrastructure/Services/DailyCheckPingService.cs
@@ -15,16 +15,20 @@ namespace DevBetterWeb.Infrastructure.Services
   {
     private TimeSpan TWO_DAYS = new TimeSpan(2, 0, 0, 0);
     private TimeSpan FOUR_DAYS = new TimeSpan(4, 0, 0, 0);
+    private int DAYS_IN_TWO_YEARS = 365 * 2;
 
-    public IRepository<Invitation> _repository;
+    public IRepository<Invitation> _inviteRepository;
+    public IRepository<Member> _memberRepository;
     public IEmailService _emailService;
     public UserManager<ApplicationUser> _userManager;
 
     public DailyCheckPingService(IRepository<Invitation> repository,
+      IRepository<Member> memberRepository,
       IEmailService emailService,
       UserManager<ApplicationUser> userManager)
     {
-      _repository = repository;
+      _inviteRepository = repository;
+      _memberRepository = memberRepository;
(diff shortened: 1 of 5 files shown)


src/DevBetterWeb.Infrastructure/Services/DailyCheckPingService.cs
```diff
@@ -64,11 +68,49 @@ public async Task SendPingIfNeeded(AppendOnlyStringList messages)
       }
     }
 
+    public async Task PingAdminsAboutAlmostAlumsIfNeeded(AppendOnlyStringList messages)
+    {
+      var members = await _memberRepository.ListAsync();
+      var membersToPingAdminsAbout = new List<Member>();
+
+      var daysBeforeGraduationToPing = 10;
```
ardalis, line 76: Maybe pull out to a constant as well. Eventually these things might come from a config file.

src/DevBetterWeb.Infrastructure/Services/DailyCheckPingService.cs
```diff
@@ -64,11 +68,49 @@ public async Task SendPingIfNeeded(AppendOnlyStringList messages)
...
+
+        foreach(var user in usersInAdminRole)
+        {
+          await _emailService.SendEmailAsync(user.Email, "Upcoming DevBetter Graduation", $"Ensure that the following member(s)'s subscriptions are not renewed, as they will graduate to alumni status in {daysBeforeGraduationToPing} days:\n {listOfMembersToPingAdminsAbout}");
+        }
+
+        messages.Append($"Admins have been reminded not to renew the subscriptions of the following members about to graduate: {listOfMembersToPingAdminsAbout}");
+      }
+    }
+
```
ardalis, line 108: This class is getting long and already violates SRP. If we end up adding another method to it, I would want to see it refactored so that at least the body of each of these methods was implemented by another service (or handler) and this service simply delegated work out to other services/handlers as needed (and would thus become a lot shorter).

ardalis (approved):

ardalis:
Looks good. Still a lot of null check warnings but we can keep working on those in future PRs. You might want to turn on "treat warnings as errors" in your local machine environment so you see the warnings more often/prominently.

src/DevBetterWeb.Infrastructure/Services/DailyCheckPingService.cs
ilyanaDev: ok

PR #381, opened 2021-07-21
do not merge yet - number of subscription plans check

ilyanaDev:
Do not merge until stripe is live. Not a big deal as doing so shouldn't break anything, but daily check will send a lot of warnings

### DIFF
--- a/src/DevBetterWeb.Core/Interfaces/IDailyCheckSubscriptionPlanCountService.cs
+++ b/src/DevBetterWeb.Core/Interfaces/IDailyCheckSubscriptionPlanCountService.cs
@@ -1,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace DevBetterWeb.Core.Interfaces
+{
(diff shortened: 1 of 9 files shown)


src/DevBetterWeb.Core/Interfaces/IDailyCheckSubscriptionPlanCountService.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace DevBetterWeb.Core.Interfaces
+{
+  public interface IDailyCheckSubscriptionPlanCountService
```
ardalis, line 5:
```suggestion
  public interface IDailyCheckSubscriptionPlanCountService2
```

ardalis: Ignore

ilyanaDev:
@ardalis  this should be able to be merged now, right? Since it's now using options?

PR #420, opened 2021-08-20
DevBetterWeb.UploaderApp added.

ShadyNagy:


src/DevBetterWeb.Core/Interfaces/IMarkdown.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace DevBetterWeb.Core.Interfaces
+{
+  public interface IMarkdown
```
ardalis, line 3: Interface name should describe what it does/what it is. If nothing else `IMarkdownService` at least tells me I should expect it to be a service I DI into things. `IMarkdownConverter` or `IMarkdownRenderer` might be good options as well.

src/DevBetterWeb.UploaderApp/Program.cs
```diff
@@ -0,0 +1,124 @@
...
+      }
+
+      var tokenIndex = argsList.FindIndex(x => x.ToLower() == "-t") + 1;
+      if (tokenIndex <= 0)
+      {
+        Console.WriteLine("PLease use -t [Vimeo token]");
+        return;
+      }
+
+      var folderToUpload = argsList[folderToUploadIndex];
```
ardalis, line 37: Ideally once you've gathered and validated the arguments to the program, you should create and call out to a service. This lets you better unit test the service, since it's not necessarily easy to test a console main() method.

ardalis (approved):
Minor suggestions

src/DevBetterWeb.Core/Interfaces/IMarkdown.cs
ShadyNagy: renamed to IMarkdownService

src/DevBetterWeb.UploaderApp/Program.cs
ShadyNagy: UploaderManager created.

ardalis (approved):

PR #435, opened 2021-09-07
Videos pagination

ShadyNagy:


ardalis:
You're still pulling in way too many commits per PR... Have you tried deleting your local and re-cloning?

src/DevBetterWeb.Web/wwwroot/js/videos.js
```diff
@@ -1,5 +1,48 @@
 ﻿$(document).ready(function () {
-
-   
-
+    var userRole = '@(User.IsInRole("Administrators") ? "true" : "false")';
```
ardalis, line 2: This code isn't going to execute on the server since this is just a static js file.

src/DevBetterWeb.Web/Controllers/VideosController.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+      _mapper = mapper;
+      _getAllVideosService = getAllVideosService;
+    }
+
+    [HttpPost("list")]
+    public async Task<IActionResult> ListAsync()
+    {
+      var draw = Request.Form["draw"].FirstOrDefault();      
+      var length = Request.Form["length"].FirstOrDefault();
```
ardalis, line 30: Why not use model binding for these values instead of reading from `Request.Form`?

src/DevBetterWeb.Web/Controllers/VideosController.cs
```diff
@@ -0,0 +1,51 @@
...
+
+
+namespace DevBetterWeb.Web.Controllers
+{
+  [Route("videos")]
+  [ApiController]
+  public class VideosController : Controller
+  {
+    private readonly IMapper _mapper;
+    private readonly GetAllVideosService _getAllVideosService;
```
ardalis, line 18: This is still calling out to the video provider, not using a local database, right? Why is this depending on an instance type, not an interface?

src/DevBetterWeb.Web/wwwroot/js/videos.js
ShadyNagy: I think it will work as I called it like this 
```csharp
@section scripts{
    <script src="~/js/videos.js" asp-append-version="true"></script>
}
```

ardalis: Check in the browser in dev tools.

And even if it does work, what prevents any end user from changing false to true themselves in client code?

ShadyNagy: it is working and the user can change the role in client code but the page will not load as it is protected in the backend also 
it is using to hide/show by role in client code only.

src/DevBetterWeb.Web/Controllers/VideosController.cs
ShadyNagy: added.

PR #450, opened 2021-09-15
Archive video applied

ShadyNagy:
- Archive video used.
- Uploader read the encoded date from the file.

### DIFF
--- a/src/DevBetterWeb.Core/Entities/Member.cs
+++ b/src/DevBetterWeb.Core/Entities/Member.cs
@@ -57,17 +57,17 @@ namespace DevBetterWeb.Core.Entities
     public decimal? CityLatitude { get; set; }
     public decimal? CityLongitude { get; set; }
     public List<BillingActivity> BillingActivities { get; set; } = new List<BillingActivity>();
-    public List<MemberVideo> Videos { get; private set; } = new List<MemberVideo>();
+    public List<MemberVideoProgress> Videos { get; private set; } = new List<MemberVideoProgress>();
 
-    public void AddVideo(MemberVideo video)
+    public void AddVideoProgress(MemberVideoProgress videoProgress)
     {
-      Guard.Against.Null(video, nameof(video));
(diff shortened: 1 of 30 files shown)


src/DevBetterWeb.Core/Entities/Member.cs
```diff
@@ -57,7 +57,19 @@ internal Member(string userId)
     public decimal? CityLatitude { get; set; }
     public decimal? CityLongitude { get; set; }
     public List<BillingActivity> BillingActivities { get; set; } = new List<BillingActivity>();
+    public List<MemberVideo> Videos { get; private set; } = new List<MemberVideo>();
 
+    public void AddVideo(MemberVideo video)
```
ardalis, line 62: They're not actually adding a video, but are adding some stats about videos they've watched, right?

src/DevBetterWeb.Core/Entities/MemberVideo.cs
```diff
@@ -0,0 +1,30 @@
+﻿using DevBetterWeb.Core.Interfaces;
+using DevBetterWeb.Core.SharedKernel;
+
+namespace DevBetterWeb.Core.Entities
+{
+  public class MemberVideo : BaseEntity, IAggregateRoot
```
ardalis, line 6: Maybe renamed to `MemberVideoProgress` ?

ardalis (approved):

src/DevBetterWeb.Core/Entities/MemberVideo.cs
ShadyNagy: Yes, that is better.
Done

src/DevBetterWeb.Core/Entities/Member.cs
ShadyNagy: changed to AddVideoProgress

ardalis:
Let's see how it works...

Issue #620, opened 2022-01-14
Fix Time Issue with Video Uploads


ardalis:
Looks like there is a time zone issue. This screenshot is from 12:03pm EST:

![image](https://user-images.githubusercontent.com/782127/149555549-fd7183ca-5222-4d1f-a3a2-fa24b33e4a96.png)

Note the first video shows it was added -11089 seconds ago (which is about 184 minutes or 3 hours) - it looks like it's assuming DateCreated is in EST time when it is actually in UTC.

samuelwine:
@ardalis Isn't there something wrong with the dateCreated time - meaning that since its a DateTimeOffset [here](https://github.com/DevBetterCom/DevBetterWeb/blob/80c2afd1a6510ac7d1db9c15eecb9f606a39fe63/src/DevBetterWeb.Web/Models/ArchiveVideoDto.cs#L15) shouldn't it be in your local time?
Additionally when I put that datetime into the JS console, it gives me 20:07 which also doesn't seem right?
![image](https://user-images.githubusercontent.com/15095510/149643800-d663f2dc-dcad-4b37-9629-1e66e03d8b74.png)

ardalis:
It's in *some* local time, not necessarily the user's local time. You can always convert a DateTimeOffset to UTC, and you can always convert UTC to any local time (ideally on the client, I think).

ardalis:
Also fix this: 
![image](https://user-images.githubusercontent.com/782127/154576591-a7b0df58-4aca-4e04-a02c-9cc57def03ab.png)
(note the "2022 years ago" bit

ShadyNagy:
This video (2019-02-23) has problem on created date on `the file Meta Data` we need to check.

> Also fix this: ![image](https://user-images.githubusercontent.com/782127/154576591-a7b0df58-4aca-4e04-a02c-9cc57def03ab.png) (note the "2022 years ago" bit

ardalis:
Yes that was a data issue - fixed.

PR #622, opened 2022-01-17
converting createdTime to UTC for updateVideo call

Joe-LaRue:
added conversion to UTC and an explicit cast to DateTimeOffset.  second part is probably overkill, but... i committed it before i could switch it back from testing!  I have not tested this solution for issue #620 !


ardalis:
This won't fix the current problems in production with existing videos. Probably we need to convert the DateTimeOffset to UTC in the client code or in the api it's calling to get the list of videos.

ardalis:
@ShadyNagy can you look into this and the underlying problem with negative time showing up immediately after videos are uploaded?

PR #653, opened 2022-02-07
Add testimonial and headshot

samuelwine:
Hi @ardalis please advise if you would like any amendments
Thank you!


ardalis:
Awesome, thanks!

PR #659, opened 2022-02-08
fix the thump startup.

ShadyNagy:


src/DevBetterWeb.Web/ConfigureServicesExtensions.cs
```diff
@@ -27,10 +27,10 @@ public static void AddStripeServices(this IServiceCollection services, string st
 
   public static void AddDailyCheckServices(this IServiceCollection services)
   {
-    services.AddHostedService<DailyCheckService>();
     services.AddScoped<IDailyCheckPingService, DailyCheckPingService>();
     services.AddScoped<IDailyCheckSubscriptionPlanCountService, DailyCheckSubscriptionPlanCountService>();
     services.AddScoped<IVideosThumbnailService, VideosThumbnailService>();
+    services.AddHostedService<DailyCheckService>();
```
ardalis, line 33: I don't see how this is changing anything? The order of service registration shouldn't change the system behavior, right?

ShadyNagy: yes you are correct

PR #711, opened 2022-05-09
Update packages and GitHub actions

KyleMcMaster:
- updates minor/patch versions of most packages 
- does not update `SixLabors.ImageSharp` since I wasn't completely aware of side effects from this
- update codeql-analysis.yml to use v2 actions


ardalis:
Thanks! Hopefully everything works but if not I'll fix it.

PR #813, opened 2022-07-13
Invitations added.

### LINKED ISSUE #708: Add Manage Invitations Page
Add a link to the page to the Admin menu.
The page should list invites.
By default it should show only active/outstanding invites, but you should be able to view used/old invites as well.
Active invites should have a button that, when clicked, deactivates that invitation.

ShadyNagy:
- No DB changes
- Fix #708

### DIFF
--- a/src/DevBetterWeb.Web/Pages/Admin/Invitations/Index.cshtml.cs
+++ b/src/DevBetterWeb.Web/Pages/Admin/Invitations/Index.cshtml.cs
@@ -1,0 +1,35 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using AutoMapper;
+using DevBetterWeb.Core;
+using DevBetterWeb.Core.Entities;
+using DevBetterWeb.Core.Interfaces;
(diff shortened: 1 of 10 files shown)


src/DevBetterWeb.Web/Pages/Admin/Invitations/Index.cshtml.cs
```diff
@@ -0,0 +1,35 @@
...
+
+  public IndexModel(IRepository<Invitation> invitationRepository, IMapper mapper)
+  {
+    _invitationRepository = invitationRepository;
+    _mapper = mapper;
+  }
+
+  public async Task<IActionResult> OnGetAsync()
+  {
+	  var invitations = await _invitationRepository.ListAsync();
```
ardalis, line 30: I might want a way to filter to just active invites but we'll stick with this for now while I see how it works.

ShadyNagy: yes, it has a filter same as I did on members.
image

PR #814, opened 2022-07-13
Shady nagy/member management

### LINKED ISSUE #407: Member management
Admins should have a way to view all members, whether active or not, and make simple changes to their details.

A member details view should include things like current subscriptions, associated invitations (which could be deactivated), whether the member is active, etc.

ShadyNagy:
- No DB changes.
Fix #407


ardalis (approved):
Hopefully there's not too much duplicate code here between the member's self-edit page and this admin page.

PR #822, opened 2022-07-14
Shady nagy/add date created admin

### LINKED ISSUE #144: Add DateCreated to User Admin Screen
Make it easy to sort by when users were added on this page:

https://devbetter.com/Admin

ShadyNagy:
- `Need` DB changes identity only
Fix #144


ardalis (approved):

ardalis:
Note: migrations don't work so I added the column by hand. I think the issue may be that the AspNetUsers type isn't part of my solution, but not sure.

ShadyNagy:
I think At some point, we need to make PR for migration

ShadyNagy:
> Note: migrations don't work so I added the column by hand. I think the issue may be that the AspNetUsers type isn't part of my solution, but not sure.

Fixed on PR #842

PR #826, opened 2022-07-18
email added to the member.

### LINKED ISSUE #311: Add `MemberByEmailSpec` Or `MemberLookupService` class with `MemberByEmail`  method
Is there no `MemberByEmailSpec`? If not, would it make sense to try and add one so you don't need the double lookup here to get a userId by email and then a member by userid?

_Originally posted by @ardalis in https://github.com/DevBetterCom/DevBetterWeb/pull/309#discussion_r640855733_

ShadyNagy:
- Need DB changes.

Fix #311
Done but is that correct?
```csharp
private async Task<Member> CreateNewMemberAsync(string userId, string firstName, string lastName, string email)
  {
    Member member = await _memberRegistrationService.RegisterMemberAsync(userId);
    member.UpdateName(firstName, lastName);
		member.UpdateEmail(email);

    return member;
  }
```
on the new member no need to update the name?


ShadyNagy:
`Hold` as I need to make a merge code to update all member email from the user.

ardalis:
I'm holding on this - let me know when you think it's ready. The prior PRs have been merged.

ShadyNagy:
It is ready to be merged.

PR #865, opened 2022-08-16
book uploaded by

### LINKED ISSUE #335: Alumni Members can add one book
As an Alumni member you should be able to add one book to the book list used for the leaderboard.

Books should list who added them - either Steve or an Alumni member - in the leaderboard and other lists.

ShadyNagy:
Fix #335 
It needs DB changes.
- Almuni can add one book from his profile.
- Leaderboard will show the book added by who.


src/DevBetterWeb.Core/Entities/Book.cs
```diff
@@ -10,11 +10,13 @@ public class Book : BaseEntity, IAggregateRoot
...
   public string? Title { get; set; }
   public string? Author { get; set; }
-  public string? Details { get; set; }
+	public int? MemberWhoUploadId { get; set; }
+	public string? Details { get; set; }
   public string? PurchaseUrl { get; set; }
   public int? BookCategoryId { get; set; }
   public int Rank { get; private set; }
 	public List<Member>? MembersWhoHaveRead { get; set; } = new List<Member>();
+	public Member? MemberWhoUpload { get; set; }
```
ardalis, line 19: You shouldn't need both an Id and a Member property. If you have Member you can always get MemberWhoUpload.Id, right?

src/DevBetterWeb.Infrastructure/Handlers/DiscordLogMemberAddBookAddHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+  private readonly BookDiscussionWebhook _webhook;
+
+  public DiscordLogMemberAddBookAddHandler(BookDiscussionWebhook webhook)
+  {
+    _webhook = webhook;
+  }
+
+  public static string returnWebhookMessageString(MemberAddedBookAddEvent domainEvent)
+  {
+    return $"User {domainEvent.Member.FirstName} {domainEvent.Member.LastName} just finished adding {domainEvent.Book.Title}! " +
```
ardalis, line 19: "User just added {title} to our book list!" instead of "just finished adding"

ardalis (approved):

src/DevBetterWeb.Infrastructure/Handlers/DiscordLogMemberAddBookAddHandler.cs
ShadyNagy: changed in next PR.

src/DevBetterWeb.Core/Entities/Book.cs
ShadyNagy: I made this for config and the EF migration but I think it is will create it automatically so yes I will remove it from here (id).

PR #906, opened 2022-09-08
Pvaillancourt/151 member birthday

### LINKED ISSUE #151: Capture Member Birthday
Add a birthday field to the member profile. It can just include month and day - we don't need year. It would be good to celebrate members' birthdays (and membership anniversary dates).

snowfrogdev:
**REQUIRES DB UPDATES**

A member can now add his birthday to his profile (day and month only).
closes #151

Birthday

### DIFF
--- a/src/DevBetterWeb.Core/ValueObjects/Birthday.cs
+++ b/src/DevBetterWeb.Core/ValueObjects/Birthday.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using CSharpFunctionalExtensions;
+
+namespace DevBetterWeb.Core.ValueObjects;
+
+public class Birthday : ValueObject
+{
+  public int Day { get; }
+  public int Month { get; }
+  private const int LEAP_YEAR = 2020;
+
+  public Birthday(int day, int month)
+  {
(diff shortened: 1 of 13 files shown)


src/DevBetterWeb.Core/ValueObjects/Birthday.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Collections.Generic;
+using CSharpFunctionalExtensions;
+
+namespace DevBetterWeb.Core.ValueObjects;
+
+public class Birthday : ValueObject
+{
+  public int Day { get; }
+  public int Month { get; }
+  private const int LEAP_YEAR = 2020;
```
ardalis, line 11: Why is this hard-coded to 2020?

snowfrogdev: I wanted to make sure that users could input 29 as a date in February. In order for the DateOnly constructor to not throw, the year has to be a leap year. 2020 is a leap year. I'm only using the DateOnly as an easy way to validate day/month combos and format them. I actually just store the day and month as ints.

If you are wondering why use const, I dunno, because it's a constant. Isn't that what const is for? It's my Typescript that is bleeding into my C#. 😁

PR #908, opened 2022-09-08
Pvaillancourt/897 crash when adding a new book

### LINKED ISSUE #897: Crash when adding a new book as an alumni
I tried adding a book and added a long multi-paragraph text to the description field and that led to an error. When I tried again but without any description at all, it worked. 

image

snowfrogdev:
**REQUIRES DB UPDATE**

- Updated `Books` `Details` column in DB to allow for up to 2000 chars as I thought 1000 was a bit limiting
- Add string length validation to the `BookViewModel` used both in Admin and MyProfile
- Converted the Details `<input>` field to a `<textarea>` for better UX

closes #897 

image

image

### DIFF
--- a/src/DevBetterWeb.Web/Pages/Admin/Books/BookViewModel.cs
+++ b/src/DevBetterWeb.Web/Pages/Admin/Books/BookViewModel.cs
@@ -1,4 +1,5 @@
 ﻿using System.ComponentModel.DataAnnotations;
+using static DevBetterWeb.Core.Constants;
 
 namespace DevBetterWeb.Web.Pages.Admin.Books;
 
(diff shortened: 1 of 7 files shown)


src/DevBetterWeb.Web/Pages/Admin/Books/BookViewModel.cs
```diff
@@ -19,6 +19,7 @@ public class BookViewModel
 	[Display(Name = "Book Category")]
 	public int? BookCategoryId { get; set; }
 
+	[StringLength(2000, ErrorMessage = "Details cannot be longer than 2000 characters.")]
```
ardalis, line 22: Might be able to use a constant, like MAX_BOOK_DESCRIPTION_LENGTH, so you don't have 2000 in quite so many places.

ardalis (commented):
See if you can remove some duplication around the 2000 value, between the attribute and the entity config.

Issue #196, opened 2021-02-04
Indexing archived videos


justSteve:
<< We really need a better way to index the videos. It's on the list, along with getting them hosted in the web app via Vimeo>>

It's an idea I've been mulling for a bit -- there are a number of 'speech-to-text' (post-event processing) services and options that, if we could script and automate the transcription process, Ardalis has a wealth of knowledge wrapped up in DevBetter's archive. Turning that video content to text content would seem like a nature first step to a more robust annotation.

I know that Office's Word app has a speech to text processor that's a) free as its included in most O365 subs; b) good enough to produce text where different speakers are differentiated and noted. Azure has something that (like Word's I've only read about.

Speaking as a new comer to the $$ group I'd really value being able get caught up on various threads of discussion but video is simply too slow. Anything that turns video to text is going to be a huge step in that direction.

I'd suggest we ask for responses here from anyone listening who has any prior experience with speech-to-text (in whatever form) share whatever lessons learned.

Thanks!

ardalis:
anyone have any thoughts on how to add transcriptions to the videos in an automated fashion? Is that something we can leverage our video hosting provider (Vimeo) to do, or would it need to be a separate custom process of ours?

This seems to suggest we can get Vimeo to do this automatically:
https://vimeo.com/blog/post/how-to-transcribe-a-video/

snowfrogdev:
> anyone have any thoughts on how to add transcriptions to the videos in an automated fashion? Is that something we can leverage our video hosting provider (Vimeo) to do, or would it need to be a separate custom process of ours?
> 
> This seems to suggest we can get Vimeo to do this automatically: https://vimeo.com/blog/post/how-to-transcribe-a-video/

The link you posted says that if you are on a paid plan, video transcription is done automatically by default. Are we on a paid plan?

snowfrogdev:
If the transcription/caption is automatic on upload, which it seems to be, it looks like we should be able to download it using Vimeo's API. See: https://developer.vimeo.com/api/reference/videos#get_text_tracks

ardalis:
Yes, we're on a paid plan. @ShadyNagy has the most experience with our Vimeo integration and their APIs if you have questions.

snowfrogdev:
I noticed that the CCs are only available in videos starting in May 2022. I'm thinking of first implementing this feature so that newly uploaded videos have their transcripts added to the video page, at the bottom. 

[...]

PR #926, opened 2022-09-14
Make transcripts awesome

snowfrogdev:
- Transcript text is now fully clickable and linked to the video details page at the appropriate timestamps corresponding to the clicked piece of text.
- I've applied a crude strategy for formatting the transcript text into paragraphs. It's not perfect, but is a heck of a lot better than a gigantic wall of text.
- Refactored to reduce the number of constructor dependencies in the Videos.Details class.
ezgif-5-5621d5b6c6

Work towards #196


src/DevBetterWeb.Web/Services/VideoDetailsService.cs
```diff
@@ -0,0 +1,63 @@
...
+using DevBetterWeb.Core.Interfaces;
+using DevBetterWeb.Core.Specs;
+using DevBetterWeb.Infrastructure.Identity.Data;
+using DevBetterWeb.Vimeo.Models;
+using DevBetterWeb.Vimeo.Services.VideoServices;
+using Microsoft.AspNetCore.Identity;
+
+namespace DevBetterWeb.Web.Services;
+
+public class VideoDetailsService
```
ShadyNagy, line 14: I suggest making an interface for this service.

src/DevBetterWeb.Web/Services/WebVTTParsingService.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace DevBetterWeb.Web.Services;
+
+public class WebVTTParsingService
```
ShadyNagy, line 8: I suggest making an interface for this service.

src/DevBetterWeb.Web/Startup.cs
```diff
@@ -114,6 +115,10 @@ public void ConfigureServices(IServiceCollection services)
 
     services.AddScoped<IVideosService, VideosService>();
     services.AddSingleton<IVideosCacheService, VideosCacheService>();
+
+    services.AddScoped<WebVTTParsingService>();
+    services.AddScoped<VideoDetailsService>();
```
ShadyNagy, line 120: If interfaces are applied.
```suggestion
    services.AddScoped<IWebVTTParsingService, WebVTTParsingService>();
    services.AddScoped<IVideoDetailsService, VideoDetailsService>();
```

src/DevBetterWeb.Web/Services/VideoDetailsService.cs
```diff
@@ -0,0 +1,63 @@
...
+using Ardalis.ApiClient;
+using DevBetterWeb.Core.Entities;
+using DevBetterWeb.Core.Interfaces;
+using DevBetterWeb.Core.Specs;
+using DevBetterWeb.Infrastructure.Identity.Data;
+using DevBetterWeb.Vimeo.Models;
+using DevBetterWeb.Vimeo.Services.VideoServices;
+using Microsoft.AspNetCore.Identity;
+
+namespace DevBetterWeb.Web.Services;
```
ShadyNagy, line 12: I think this service needs to be in the Core project.

src/DevBetterWeb.Web/Services/WebVTTParsingService.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace DevBetterWeb.Web.Services;
```
ShadyNagy, line 6: I think this service needs to be in the Core project.

ShadyNagy (changes requested):
Some suggestions are created.

snowfrogdev:
@ShadyNagy @ardalis I think I would normally consider the WebVTTParsingService and VideoDetailsService as application services, not domain services. 

The WebVTTParsingService can only serve a web client, it takes a .vtt formated string and produces HTML, it doesn't use any of our domain entities and it doesn't deal with persistence. As such, I feel like it shouldn't be in the Core but in Web. That service could have very easily been implemented in JS and run directly on the client. What do you think?

As for the VideoDetailsService I would also not consider it a Domain Service, as it doesn't have domain business logic and merely serves as a direct client to the domain model. I would tend to put Domain Services in the Core and Application Services outside the Core. But that may not be the way you do things. Let me know what you think.

Finally, in terms of the interfaces, I always make interfaces for Domain Services, but I often feel like it's superfluous in the case of Application Services. What's your thinking on this? What benefit, vs the cost of adding an abstraction, do we get from adding an interface to our Application Service? Would you normally unit test an application service? I would think they're probably more likely to be part larger component/integration level tests. Thoughts?

ardalis:
Just the name of the WebVTTParsingService makes me think it doesn't really belong in Core, because it is so implementation detail specific. So, no argument there. VideoDetailsService certainly sounds more generic and high level, so I think a case might be made either way for it (based solely on its name). If you were able to make it a Domain Service (moving it into Core) it would probably make it easier to unit test (or would at least advertise that it was easy to unit test by virtue of its being in Core). I don't tend to make a big distinction between domain and application services aside from the types they deal with. If the types used are UI types (i.e. it returns a viewmodel or API DTO rather than a domain object), then I tend to classify it as an application service and put it in the UI/Web project. If the types used are all domain types, then I'm more likely to try to put it in Core/domain model.

As for interfaces, they rarely hurt and often help. Having an interface makes it easier to flexibly apply patterns like Decorator to the service, and this is often useful for caching and other behavior-adding cross-cutting concerns. When I see a constructor that's taking in a concrete type rather than an interface, I consider that a code smell. It's not wrong, but I wonder why the author chose to tightly couple to a specific implementation rather than allowing for flexibility (such as decorators, etc.) by coupling to an interface instead. If I err on one side or the other with interfaces in C#, I'm sure I err on the side of over, not under, using them.

src/DevBetterWeb.Web/Services/WebVTTParsingService.cs
snowfrogdev: Base on @ardalis' [comments](https://github.com/DevBetterCom/DevBetterWeb/pull/926#issuecomment-1246997543) I will keep the service in the Web project.

src/DevBetterWeb.Web/Services/VideoDetailsService.cs
snowfrogdev: Because this service depends on the Infrastructure and Vimeo projects, I will keep it in the  Web project for the moment.

ShadyNagy (approved):
It is perfect.
Maybe we need to think later on this
```csharp
Task<(HttpResponse<Video>, HttpResponse<GetAllTextTracksResponse>, ArchiveVideo?, ApplicationUser)>
```

Issue #1025, opened 2023-02-16
WebHook suggestion


ShadyNagy:
Hi friends 
Do you like to replace the [existing WebHook](https://github.com/DevBetterCom/DevBetterWeb/blob/main/src/DevBetterWeb.Infrastructure/DiscordWebhooks/BaseWebhook.cs) with this service and totally remove the Discord package?

```csharp
public class DiscordWebhookService : IDiscordWebhookService
  {
    private readonly HttpClient _httpClient = new HttpClient();
    protected string _webhookUrl;
    private readonly IAppLogger<DiscordWebhookService> _logger;

    public DiscordWebhookService(IAppLogger<DiscordWebhookService> logger)
    {
      _logger = logger;
    }

    public void CreateUrl(string webhookUrl)
    {
      _webhookUrl = webhookUrl;
    }

    public void CreateUrl(ulong id, string token)
    {
      _webhookUrl = $"https://discordapp.com/api/webhooks/{id}/{token}";
    }

    public async Task<HttpResponseMessage> SendAsync(string contentBody, string username = "", string avatarUrl = "", bool isTTS = false, IEnumerable<Embed> embeds = null)
    {
      var request = new WebHookRequest(contentBody, username, avatarUrl, isTTS, embeds);
      var content = new StringContent(JsonSerializer.Serialize(request), Encoding.UTF8, "application/json");
      try
      {
        return await _httpClient.PostAsync(_webhookUrl, content);
      }
      catch (Exception ex)
      {
        _logger.LogError(ex, "Error sending webhook", _webhookUrl, content);
        return new HttpResponseMessage();
      }
    }
  }
```

ShadyNagy:
Discord models
```csharp
public class Embed : IEmbedUrl
  {
    [JsonPropertyName("title")]
    public string Title { get; set; }
    [JsonPropertyName("type")]
    public string Type { get; set; } = "rich";
    [JsonPropertyName("description")]
    public string Description { get; set; }
    public string Url { get; set; }
    [JsonPropertyName("timestamp")]
    public DateTimeOffset? TimeStamp { get; set; }
    [JsonPropertyName("color")]
    public int Color { get; set; }
    [JsonPropertyName("footer")]
    public EmbedFooter Footer { get; set; }
    [JsonPropertyName("image")]
    public EmbedImage Image { get; set; }
    [JsonPropertyName("thumbnail")]
    public EmbedThumbnail Thumbnail { get; set; }
[...]

PR #1036, opened 2023-03-10
discord refactor

ShadyNagy:
closes issue #1025

### DIFF
--- a/src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
+++ b/src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
@@ -32,7 +32,7 @@ public class DiscordWebhookService : IDiscordWebhookService
 		var request = new WebHookRequest(contentBody, username, avatarUrl, isTTS, embeds);
 		var json = JsonSerializer.Serialize(request);
 
-		var content = new StringContent(json, Encoding.UTF8, "application/json");
+		using var content = new StringContent(json, Encoding.UTF8, "application/json");
 		try
 		{
 			return await _httpClient.PostAsync(webhookUrl, content);
--- a/src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
+++ b/src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
@@ -12,7 +12,7 @@ namespace DevBetterWeb.Infrastructure.Services;
 
 public class DiscordWebhookService : IDiscordWebhookService
 {
-	private readonly HttpClient _httpClient = new HttpClient();
+	private readonly HttpClient _httpClient = new();
 	private readonly IAppLogger<DiscordWebhookService> _logger;
 
 	public DiscordWebhookService(IAppLogger<DiscordWebhookService> logger)
--- a/src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
+++ b/src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
@@ -35,7 +35,10 @@ public class DiscordWebhookService : IDiscordWebhookService
 		using var content = new StringContent(json, Encoding.UTF8, "application/json");
 		try
 		{
-			return await _httpClient.PostAsync(webhookUrl, content);
+			var response = await _httpClient.PostAsync(webhookUrl, content);
+			response.EnsureSuccessStatusCode();
+
+			return response;
 		}
 		catch (Exception ex)
 		{
--- a/src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
+++ b/src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
@@ -1,0 +1,46 @@
+﻿using DevBetterWeb.Infrastructure.DiscordWebhooks;
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+using DevBetterWeb.Core.Interfaces;
+using DevBetterWeb.Infrastructure.Interfaces;
+
+namespace DevBetterWeb.Infrastructure.Services;
+
+public class DiscordWebhookService : IDiscordWebhookService
+{
+	private readonly HttpClient _httpClient = new HttpClient();
+	private readonly IAppLogger<DiscordWebhookService> _logger;
+
+	public DiscordWebhookService(IAppLogger<DiscordWebhookService> logger)
+	{
+		_logger = logger;
+	}
+
+	public Task<HttpResponseMessage> SendAsync(ulong id, string token, string contentBody, string? username = null, string? avatarUrl = null, bool isTTS = false, IEnumerable<Embed>? embeds = null)
+	{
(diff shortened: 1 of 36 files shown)


ardalis (approved):

snowfrogdev (approved):
I think this is definitely an improvement and a step in the right direction. Introducing the `DiscordWebhookService` and having that be in the DI instead of only the Webhooks themselves is much better.

But looking at the whole system, I'm left wondering if we can't simplify the whole thing. I guess I feel some unease at the amount of indirection and decoupling we have for the sole purpose of sending messages to our Discord server. 

I don't know if there is necessarily a better approach but it just feels like so much code, so many classes, for one simple operation that only changes along two axis (webhook url, message).

For example, I can't help but feel that `BaseWebhook`, and its derived classes, could be completely eliminated. We have a set of Handlers called `DiscordLog[EventName]Handler` which, I gather from their name, have the sole purpose of sending messages to Discord. We could have those handlers depend directly on our new `DiscordWebhookService` and `IOptions<DiscordWebhookUrls>`. It would be their only two dependencies, which hardly seems unreasonable.

src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
```diff
@@ -0,0 +1,46 @@
...
+	public async Task<HttpResponseMessage> SendAsync(string webhookUrl, string contentBody, string? username = null, string? avatarUrl = null, bool isTTS = false, IEnumerable<Embed>? embeds = null)
+	{
+		var request = new WebHookRequest(contentBody, username, avatarUrl, isTTS, embeds);
+		var json = JsonSerializer.Serialize(request);
+
+		using var content = new StringContent(json, Encoding.UTF8, "application/json");
+		try
+		{
+			return await _httpClient.PostAsync(webhookUrl, content);
+		}
```
snowfrogdev, line 39: @ShadyNagy How are we handling the case where we are not getting a success here?

snowfrogdev (changes requested):

src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
snowfrogdev: And why are we returning the Response if we never do anything with it?

src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
```diff
@@ -0,0 +1,46 @@
...
+	{
+		_logger = logger;
+	}
+
+	public Task<HttpResponseMessage> SendAsync(ulong id, string token, string contentBody, string? username = null, string? avatarUrl = null, bool isTTS = false, IEnumerable<Embed>? embeds = null)
+	{
+		var webhookUrl = $"https://discordapp.com/api/webhooks/{id}/{token}";
+
+		return SendAsync(webhookUrl, contentBody, username, avatarUrl, isTTS, embeds);
+	}
```
snowfrogdev, line 28: I believe this to be dead code; not being used or called anywhere. Judging by the pattern we have set, it is unlikely to ever be use. Don't forget YAGNI and KISS 😉.

src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
ShadyNagy: > And why are we returning the Response if we never do anything with it?

I think it is the same as what you did in your code `return response;` ! ;) This also need to use using in `var content = new StringContent(JsonConvert.SerializeObject(this), Encoding.UTF8, "application/json");`
```csharp
public virtual async Task<HttpResponseMessage> Send()
	{
		var content = new StringContent(JsonConvert.SerializeObject(this), Encoding.UTF8, "application/json");
		try
		{
			var response = await _httpClient.PostAsync(_webhookUrl, content);
			var responseContent = await response.Content.ReadAsStringAsync();
			response.EnsureSuccessStatusCode();
			return response;
		}
		catch (Exception ex)
		{
			_logger.LogError(ex, "Error sending webhook", _webhookUrl, content);
			return new HttpResponseMessage();
		}
	}
```

src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
ShadyNagy: > I believe this to be dead code; not being used or called anywhere. Judging by the pattern we have set, it is unlikely to ever be used. Don't forget YAGNI and KISS 😉.

I already separated this into another project and take it from there. I think we can do the same and release it on NuGet later.
The same concept you are using here 
```csharp
public BaseWebhook(ulong id, string token, IAppLogger<BaseWebhook> logger) : this($"https://discordapp.com/api/webhooks/{id}/{token}", logger)
	{
	}
```

src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
ShadyNagy: > @ShadyNagy How are we handling the case where we are not getting a success here?

EnsureSuccessStatusCode is added

ShadyNagy:
> I think this is definitely an improvement and a step in the right direction. Introducing the `DiscordWebhookService` and having that be in the DI instead of only the Webhooks themselves is much better.
> 
> But looking at the whole system, I'm left wondering if we can't simplify the whole thing. I guess I feel some unease at the amount of indirection and decoupling we have for the sole purpose of sending messages to our Discord server.
> 
> I don't know if there is necessarily a better approach but it just feels like so much code, so many classes, for one simple operation that only changes along two axis (webhook url, message).
> 
> For example, I can't help but feel that `BaseWebhook`, and its derived classes, could be completely eliminated. We have a set of Handlers called `DiscordLog[EventName]Handler` which, I gather from their name, have the sole purpose of sending messages to Discord. We could have those handlers depend directly on our new `DiscordWebhookService` and `IOptions<DiscordWebhookUrls>`. It would be their only two dependencies, which hardly seems unreasonable.

You will repeat yourself many times.

src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
snowfrogdev: I'm not sure I understand, I'm sorry. Could you help me understand how this method is being used in devBetter at the moment?

src/DevBetterWeb.Infrastructure/Services/DiscordWebhookService.cs
snowfrogdev: If we're not using the response we might as well not return it from the function. This way we'll reduce possible confusion in future readers.

And do you think we can actually log the response content if ever the request is not successful? It will help with diagnostics if ever there's a problem.

snowfrogdev:
> You will repeat yourself many times.

Not much more than now, I think. I feel like what little repetition may be introduced is worth it considering it allows us to remove an entire layer of abstraction and indirection that, I feel, provides little value.

My analysis could be wrong though. @ardalis @KyleMcMaster, any thoughts that may help steer our design decision here?

ardalis:
Thanks, Shady! Merging and we can keep iterating separately if needed. I don't have a strong opinion without digging into the code more.

PR #1093, opened 2023-05-10
1092 book details remove non members fix

### LINKED ISSUE #1092: Book details display non members
image

ShadyNagy:
## Description
In this Pull Request, we address an issue in the BookDetailsModel of our application. Previously, the book details page was displaying all members who have read a particular book, including non-current members. This was not the intended behavior.

With this PR, we have updated the OnGet method within the BookDetailsModel to correctly fetch and display only the current members who have read the book. This is done by making use of our ILeaderboardService's GetBookDetailsAsync method to fetch the book's details, which now correctly filters out non-current members.

## Changes

- Added a new method `GetBookDetailsAsync` in `IFilteredBookDetailsService`. This method takes a `bookId` as a parameter and returns a `BookDetailsViewModel`. It fetches the book details using the provided book ID and then filters out the non-current members before returning the result.

- Implemented `GetBookDetailsAsync` in `FilteredBookDetailsService`. It uses `IBookService` to fetch the book details and then `IFilteredLeaderboardService` to filter out non-current members from the book details.

- Created a new `BookService` implementing `IBookService`. It has a `GetBookByIdAsync` method that fetches the book details from the repository and constructs a `BookDetailsViewModel`.

- Added a `RemoveNonCurrentMembersFromBookDetailsAsync` method to `IFilteredBookDetailsService`. This method takes a `BookDetailsViewModel` and removes non-current members from it. The implementation of this method is in `FilteredBookDetailsService`.

- Updated `BookDetailsModel` in the book details screen to use the new `GetBookDetailsAsync` method from `IBookDetailsService`. This ensures that the book details shown on the screen will only include current members.

## Tests

We have thoroughly tested the new implementations with a series of unit tests. Here's the overview:

1. **RemoveNonCurrentMembersFromBookDetailsAsync**

   - `RemoveNonCurrentMembersGivenNonCurrentMembersInBookDetailsViewModel()`
     - This test ensures that the method correctly removes non-current members from a `BookDetailsViewModel` instance.

   - `NotRemoveCurrentMembersGivenCurrentMembersInBookDetailsViewModel()`
     - This test makes sure that current members are not removed.

   - `ReturnSameBookDetailsGivenNoNonCurrentMembersInBookDetailsViewModel()`
     - This test confirms that the original `BookDetailsViewModel` is returned when there are no non-current members.

   - `HandleEmptyMembersListGivenBookDetailsViewModelWithNoMembers()`
     - This test verifies that the method can handle an empty members list without throwing exceptions.

   - `RemoveAllMembersGivenAllNonCurrentMembersInBookDetailsViewModel()`
     - This test ensures that all members are removed when they are all non-current.

2. **GetBookDetailsAsync**

   - `ReturnsNullGivenNonExistentBookId()`
     - This test confirms that the method correctly returns null when provided with a non-existent book ID.

   - `ReturnBookDetailsGivenValidBookId()`
     - This test verifies that the method returns the correct `BookDetailsViewModel` when provided with a valid book ID.

   - `FilterNonCurrentMembersGivenValidBookId()`
     - This test ensures that non-current members are correctly filtered out of the `BookDetailsViewModel` when provided with a valid book ID.

## Impact
This change impacts the user interface on the book details page.

## Checklist
- [x] Coding the services
- [x] Code compiles correctly
- [x] Created tests
- [x] All tests passing

## Related Issue
closes #1092

### DIFF
--- a/src/DevBetterWeb.Web/Interfaces/ILeaderboardService.cs
+++ b/src/DevBetterWeb.Web/Interfaces/ILeaderboardService.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Threading.Tasks;
 using DevBetterWeb.Web.Models;
+using DevBetterWeb.Web.Pages.Leaderboard;
 
 namespace DevBetterWeb.Web.Interfaces;
 
@@ -10,4 +11,5 @@ namespace DevBetterWeb.Web.Interfaces;
 public interface ILeaderboardService
 {
 	Task<List<BookCategoryDto>> SetBookCategoriesAsync();
+	Task<BookDetailsViewModel?> GetBookDetailsAsync(string bookId);
 }
--- a/src/DevBetterWeb.Web/Interfaces/ILeaderboardService.cs
+++ b/src/DevBetterWeb.Web/Interfaces/ILeaderboardService.cs
@@ -11,5 +11,4 @@ namespace DevBetterWeb.Web.Interfaces;
 public interface ILeaderboardService
 {
 	Task<List<BookCategoryDto>> SetBookCategoriesAsync();
-	Task<BookDetailsViewModel?> GetBookDetailsAsync(string bookId);
 }
--- a/src/DevBetterWeb.Web/Pages/Leaderboard/BookDetails.cshtml.cs
+++ b/src/DevBetterWeb.Web/Pages/Leaderboard/BookDetails.cshtml.cs
@@ -1,7 +1,5 @@
 ﻿using System.Threading.Tasks;
-using DevBetterWeb.Core.Entities;
-using DevBetterWeb.Core.Interfaces;
-using DevBetterWeb.Core.Specs;
+using DevBetterWeb.Web.Interfaces;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
@@ -11,22 +9,18 @@ namespace DevBetterWeb.Web.Pages.Leaderboard;
 [Authorize]
 public class BookDetailsModel : PageModel
 {
-  public BookDetailsViewModel? BookDetailsViewModel { get; set; }
-  private readonly IRepository<Book> _bookRepository;
+	private readonly ILeaderboardService _leaderboardService;
+	public BookDetailsViewModel? BookDetailsViewModel { get; set; }
 
-  public BookDetailsModel(IRepository<Book> bookRepository)
+	public BookDetailsModel(ILeaderboardService leaderboardService)
   {
-    _bookRepository = bookRepository;
+	  _leaderboardService = leaderboardService;
   }
 
   public async Task<IActionResult> OnGet(string bookId)
   {
-    var spec = new BookByIdWithMembersSpec(int.Parse(bookId));
-    var book = await _bookRepository.FirstOrDefaultAsync(spec);
+    BookDetailsViewModel = await _leaderboardService.GetBookDetailsAsync(bookId);
 
-    if (book == null) return NotFound(bookId);
-
-    BookDetailsViewModel = new BookDetailsViewModel(book!);
-    return Page();
+		return Page();
   }
 }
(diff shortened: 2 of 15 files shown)


ardalis:
Nice pull request template! Where'd that come from?

src/DevBetterWeb.Web/Interfaces/ILeaderboardService.cs
```diff
@@ -10,4 +11,5 @@ namespace DevBetterWeb.Web.Interfaces;
 public interface ILeaderboardService
 {
 	Task<List<BookCategoryDto>> SetBookCategoriesAsync();
+	Task<BookDetailsViewModel?> GetBookDetailsAsync(string bookId);
```
ardalis, line 14: Why is this on a leaderboard service? The list of members who have read a book isn't a leaderboard, it it?

src/DevBetterWeb.Web/Services/BookService.cs
```diff
@@ -0,0 +1,40 @@
...
+	/// Initializes a new instance of the <see cref="BookService"/> class.
+	/// </summary>
+	/// <param name="bookRepository">The book category repository.</param>
+	public BookService(IRepository<Book> bookRepository)
+	{
+		_bookRepository = bookRepository;
+	}
+
+
+	public async Task<BookDetailsViewModel?> GetBookByIdAsync(int bookId)
```
ardalis, line 30: This is more what I would expect the book details page to call, a `BookService`...

src/DevBetterWeb.Web/Pages/Leaderboard/BookDetails.cshtml.cs
```diff
@@ -11,22 +9,18 @@ namespace DevBetterWeb.Web.Pages.Leaderboard;
...
   {
-    _bookRepository = bookRepository;
+	  _leaderboardService = leaderboardService;
   }
 
   public async Task<IActionResult> OnGet(string bookId)
   {
-    var spec = new BookByIdWithMembersSpec(int.Parse(bookId));
-    var book = await _bookRepository.FirstOrDefaultAsync(spec);
+    BookDetailsViewModel = await _leaderboardService.GetBookDetailsAsync(bookId);
```
ardalis, line 22: Why isn't this calling `IBookService`?

ardalis (changes requested):
Not sure why the leaderboardservice is being used...

src/DevBetterWeb.Web/Interfaces/ILeaderboardService.cs
ShadyNagy: @ardalis 
So is better to make a new service for book details or just add this code to the book details page
```csharp
	private async Task<BookDetailsViewModel?> GetBookDetailsAsync(string bookId)
	{
		var bookDetails = await _bookService.GetBookByIdAsync(int.Parse(bookId));
		if (bookDetails == null)
		{
			return null;
		}
		bookDetails = await _filteredLeaderboardService.RemoveNonCurrentMembersFromBookDetailsAsync(bookDetails);

		return bookDetails;
	}
```

ShadyNagy:
@ardalis 
So is better to make new service for book details or just use this code in the book details page
```csharp
	private async Task<BookDetailsViewModel?> GetBookDetailsAsync(string bookId)
	{
		var bookDetails = await _bookService.GetBookByIdAsync(int.Parse(bookId));
		if (bookDetails == null)
		{
			return null;
		}
		bookDetails = await _filteredLeaderboardService.RemoveNonCurrentMembersFromBookDetailsAsync(bookDetails);

		return bookDetails;
	}
```
> Not sure why the leaderboardservice is being used...

ardalis:
Keep logic out of pages when you can - they're harder to test and less reusable. Put logic into services as your default approach. It's fine to have smaller, single-responsibility services (or at least well-named ones) than larger services that aren't cohesive (like a leaderboard service with a book details method on it that has nothing to do with leaderboards).

src/DevBetterWeb.Web/Interfaces/ILeaderboardService.cs
ShadyNagy: done

src/DevBetterWeb.Web/Pages/Leaderboard/BookDetails.cshtml.cs
ShadyNagy: done

src/DevBetterWeb.Web/Services/BookService.cs
ShadyNagy: done

PR #1169, opened 2023-09-29
use vimeo package

ShadyNagy:
- New vimeo package is used.
- Base folder is applied.


sadukie (changes requested):
Marking request changes to hold the PR until reviewed

sadukie:
- In looking at the functionality with Shady, we tried moving a file around in Vimeo and it looks like the video ID does not change. There shouldn't be any database changes required for this.

- We need to make sure to have the following settings for devBetter:
  - Token: Vimeo Token
  - BaseFolderId: 0 (get this from the Vimeo site)
 
 @ardalis Let's talk about this to make sure we can bring this in safely.

ardalis (commented):
accepting changes

PR #1193, opened 2023-10-24
Fixing dependency issues

sadukie:
This allows us to run locally with the mismatch in package versions.

The problem is Microsoft.Extensions.Identity.Stores uses 7.0.13 but Microsoft.Extensions.Identity.Core is at 7.0.10.  Bringing them both to 7.0.13.


sadukie:
Note: Only .Stores was updated in this dependabot PR: https://github.com/DevBetterCom/DevBetterWeb/pull/1192

ardalis:
Yeah Dependabot sometimes misses things - I wish for the framework libraries that all rev together and have interdependencies it would somehow do those as one PR...

PR #1245, opened 2024-01-05
Update avatar for upper right menu is fixed

ShadyNagy:


ardalis (approved):

ardalis:
Cool, let's see what it does...

Issue #1141, opened 2023-08-15
Replace Moq with NSubstitute


ardalis:
See: 
https://ardalis.com/porting-moq-nsubstitute/

sadukie:
See also: https://blog.nimblepros.com/blogs/moq-vs-nsubstitute-code-comparisons/

radekwojpl2:
If this task is free I can do that.

ardalis:
Go for it

radekwojpl2:
Here is pull request https://github.com/DevBetterCom/DevBetterWeb/pull/1165.
I have notice that tests for DevBetterWeb.Vimeo.Tests are not working, should I do something with them ?

ardalis:
@radekwojpl2 can you fix the tests/build and I'll pull in your PR?

PR #1272, opened 2024-03-08
#1141 Replace Moq with NSubstitute

sdepouw:
- Completes #1141 
- Removes `Moq` NuGet package and replaces it with `NSubstitute`
- Removes `ILogger.Moq` NuGet package
  - This was replaced with `NSubstituteLoggerExtensions` (custom implementation), which is open to be expanded, but was done bare-minimum style to make the tests continue to pass with verifying `ILogger` calls.
- Removed a stray floating `</a>` from a view


ardalis:
Thanks!

PR #1313, opened 2024-07-27
require address for members signing up

### LINKED ISSUE #1312: Require Address for Members Signing Up
For tax reasons we require new members' addresses. This needs to be incorporated into the new member signup flow before we run their subscription through Stripe.

ShadyNagy:
Closes #1312


src/DevBetterWeb.Core/Entities/Member.cs
```diff
@@ -123,11 +126,14 @@ public void UpdateName(string? firstName, string? lastName, bool isEvent = true)
...
 
-	public void UpdateAddress(string? address, bool isEvent = true)
+	public void UpdateAddress(string? address, string? city, string? country, string? postalCode, bool isEvent = true)
 	{
 		if (Address == address) return;
 
 		Address = address;
+		City = city;
+		Country = country;
+		PostalCode = postalCode;
```
lloydkevin, line 137: Do we need similar equality checks for the new fields like in `address`?

ShadyNagy: This checks will not work in the same way need to be inverted.

ShadyNagy:
The Shipping Address is used.

ShadyNagy:
@ardalis 
It's ready for merge.

ardalis:
🤞

PR #1329, opened 2024-08-26
add email notification when new video available solve

### LINKED ISSUE #1134: Add Discord Notification When a New Video is Available
Allow members to opt into receiving an email whenever a new video is published.

ShadyNagy:
When new video added it will send an email for users who have roles.

Closes #1134


ardalis (commented):
We can't just have this be based on user roles - they would need to also opt in or out of getting the emails. Probably we need an option in the user's profile that would say "Email me  when new video recordings are added" and then we can key off of that for this feature.

Issue #1424, opened 2026-07-19
Getting Bad Request when trying to log in


sadukie:
Saw this reported in devBetter Discord.

Confirmed with Azure logs from the App Service:

```
2026-07-19T17:07:32.6248652Z Microsoft.AspNetCore.Antiforgery.AntiforgeryValidationException: The antiforgery token could not be decrypted.
2026-07-19T17:07:32.6248686Z  ---> System.Security.Cryptography.CryptographicException: An error occurred during a cryptographic operation. Refer to the inner exception for more information.
2026-07-19T17:07:32.6248843Z  ---> System.MissingMethodException: Method not found: 'Boolean Microsoft.AspNetCore.Cryptography.CryptoUtil.TimeConstantBuffersAreEqual(Byte[], Int32, Int32, Byte[], Int32, Int32)'.
2026-07-19T17:07:32.6248878Z    at Microsoft.AspNetCore.DataProtection.Managed.ManagedAuthenticatedEncryptor.Decrypt(ArraySegment`1 protectedPayload, ArraySegment`1 additionalAuthenticatedData)
2026-07-19T17:07:32.6248901Z    --- End of inner exception stack trace ---
2026-07-19T17:07:32.6248929Z    at Microsoft.AspNetCore.DataProtection.Managed.ManagedAuthenticatedEncryptor.Decrypt(ArraySegment`1 protectedPayload, ArraySegment`1 additionalAuthenticatedData)
2026-07-19T17:07:32.6248957Z    at Microsoft.AspNetCore.DataProtection.KeyManagement.KeyRingBasedDataProtector.UnprotectCore(Byte[] protectedData, Boolean allowOperationsOnRevokedKeys, UnprotectStatus& status)
2026-07-19T17:07:32.6248983Z    at Microsoft.AspNetCore.DataProtection.KeyManagement.KeyRingBasedDataProtector.Unprotect(Byte[] protectedData)
2026-07-19T17:07:32.6249007Z    at Microsoft.AspNetCore.Antiforgery.DefaultAntiforgeryTokenSerializer.Deserialize(String serializedToken)
2026-07-19T17:07:32.624903Z    --- End of inner exception stack trace ---
2026-07-19T17:07:32.6249055Z    at Microsoft.AspNetCore.Antiforgery.DefaultAntiforgeryTokenSerializer.Deserialize(String serializedToken)
[...]

PR #1425, opened 2026-07-19
Finish .NET 10 Version Upgrades

Coop56:
Some projects and dependencies still pointed at .NET 9.  This should resolve the login issues.

PR #1426, opened 2026-07-19
Fixing antiforgery issues - finishing a .NET 10 update

sadukie:
See #1424 for more details

Once this gets committed, make sure to update the stack on the Azure App Service to be using .NET 10.


sadukie:
deferring to #1425

